PR #4, opened 2014-09-11
Use full schema.org URLs for all mappings

hubgit:
Rationales:

* Not having to deal with namespace prefixes makes things easier for parsers.
* Might as well use schema.org mappings for all the properties where they exist.

### DIFF
--- a/codemeta.jsonld
+++ b/codemeta.jsonld
@@ -1,33 +1,25 @@
 {
-   "@context":
-   {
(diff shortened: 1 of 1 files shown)


mbjones:
Discussed namespace issues at the CodeMeta workshop, made decisions that have been checked into the newly merged jsonld file.  Additional changes from the workshop will be pushed over the next few days as wel clean it all up.  Closing.

PR #27, opened 2016-03-11
Add the Software Entities Model

zoidy:
From here:
http://rrr.cs.st-andrews.ac.uk/wp-content/uploads/2015/10/guidelines-software-identification.pdf

This is a DataCite-based schema to represent software. It's slightly different than the existing column for Datacite which is why I added it.

Note that this is a partial map as some of the concepts don't appear to be present in CodeMeta.

### DIFF
--- a/crosswalk.csv
+++ b/crosswalk.csv
@@ -1,36 +1,36 @@
-Concept,CodeMeta Field,Zenodo,Figshare,code.jsonld,Software Discovery Index,Software Ontology,DataCite,Dublin Core
-SoftwareTitle,title,title,Title,name,Software title,,Title,title
-SoftwareIdentifier,identifier,id,?,citation,Persistent identifier,,Identifier,identifier
-SoftwareAuthor,author,creators,Authors(s),author,Author names and affiliations,,Creator,creator
-AuthorName, name,name,Name,name,-,,creatorName,-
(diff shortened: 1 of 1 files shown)


cboettig:
@zoidy Haven't had a chance to take a close look, but this looks great. What version of the DataCite schema is this based on?  (@mfenner has mentioned an upcoming version 4.0 they have been working on, and might have comments on this PR as well)

mfenner:
Not really any changes in the upcoming DataCite Schema 4.0 that would be relevant here. The Schema will be released in Q1 2016.

zoidy:
@cboettig @mfenner The document doesn't state specifically but it's probably 3.1. At IDCC last month, the authors [made reference](http://www.dcc.ac.uk/webfm_send/2169) to the use of the potential use of descriptionType="TechnicalInfo" for the Description field. If I'm not mistaken, TechnicalInfo is only present in 4.0

mbjones:
This looks good, @zoidy -- but it would also be useful to have the concepts that are not covered by codemeta represented as new Concepts in the table.  How many new concepts are there?  I briefly read the JISC report and it seems like there is a lot of overlap, albeit some nice conceptualizations around the model for software.  I merged this as is, but maybe you could do another PR with the additional concepts?

PR #64, opened 2016-04-17
Use minitest for more informative error states

no-reply:


acabunoc:
👍 worth it to get better error states

PR #71, opened 2016-04-17
Unit test for the email pattern

no-reply:


acabunoc:
👍 thanks @no-reply 
LGTM

PR #73, opened 2016-04-19
Add test for Agent example

no-reply:
Adds an initial test for the agent example file. Note that this test currently fails, as the file is not valid JSON.

Moves the test script into a Rake task to handle multiple minitest files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: ruby
 bundler_args: --without debug
-script: "bundle exec ruby minitest/test_example_codemeta.rb"
+script: "bundle exec rake test"
 env:
   - CI=true
 rvm:
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,6 @@
 source "https://rubygems.org"
 
+gem 'rake'
 gem 'minitest'
+
 gem 'json-ld'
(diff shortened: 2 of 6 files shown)


no-reply:
8816c98 fixes the JSON syntax that was causing the failure.

I didn't check whether the resulting graph looks reasonable.

mbjones:
Thanks, @no-reply -- I'll note that I folded the agent blocks into the main example-codemeta.json file, so the temporary example-codemeta-agent.json file is no longer needed, and we probably don't need to test it and that file can be removed at this point -- I had only checked it in to pass it onto @cboettig to work on the schema additions.  Do you think the other changes, like adding the rake dependency and rakefile, are useful even if we remove the additional test file?

cboettig:
@mbjones I think we should keep them.  From the discussion on Sunday it sounded like it would be useful to have a suite of different example-codemeta.json files, so that some could be more minimal and some represent richer documentation geared at specific use cases.

Thus the additional test infrastructure @no-reply added to test the 'agent' example separately will I think still be helpful when we want to run the tests over a suite of example files.

If it makes the repository more tidy I suppose we could move all the test infrastructure to it's own sub-directory.

mbjones:
Yeah, we need some more organization, with a directory for test files, at a minimum.  I'll go ahead and merge this PR and we can adjust after that.

no-reply:
:+1: the rake dependency and Rakefile are unlikely to add maintenance overhead and will give you magic pattern matching for any future test files.

PR #84, opened 2016-07-14
Fix small typo

nuest:


gothub:
@nuest Thanks for the fix!

Issue #105, opened 2016-09-14
Create documentation/FAQ suitable for newcomers


sdruskat:
This should probably live outside of README.md (in a separate `docs` folder?).

sdruskat:
A draft:

# CodeMeta FAQ

## What CodeMeta IS

CodeMeta is a "[Rosetta Stone](https://en.wikipedia.org/wiki/Rosetta_Stone)", i.e., a translation medium, for software metadata.

## What CodeMeta ISN'T

- CodeMeta is **not** a new standard! Instead, it is designed to **foster collaboration and interoperability** of metadata standards

## Rationale: Why CodeMeta?

CodeMeta exists to

-  **incentivise** Research Software Engineers and domain scientists developing software to **release their software into the open**
- allow research software to be **cited more easily**
- **increase discoverability** of research software
- allow people to **build tools** that use the metadata, e.g., for indexing, dependency checking, etc.

## Does CodeMeta have a format?

- There is a [JSON-LD representation](../../blob/master/codemeta.jsonld) for CodeMeta.

## What can I do with the CodeMeta format?

- **[Discoverability]** Build a searchable index or repository of [CodeMeta-compatible projects](#what-does-"codemeta-compatible"-mean?)
- **[Discovery]** Easily track and compare dependencies and licenses of [CodeMeta-compatible projects](#what-does-"codemeta-compatible"-mean?)
- **Add your idea here!** 

## What does "CodeMeta-compatible" mean?

???

# Getting started

## Create an implementation

`[TBD]` Link to the spec

## Use a CodeMeta library

`[TBD]` Is there such a thing? This section should include links to source, download, and a tutorial

## Add a new format

Please contact us if you want a new format to be added to CodeMeta.

***WHAT CONTACT DETAILS SHOULD BE LINKED TO HERE?***

# Contribute

Please either contact us, or fork the repository and create a pull request (or both).

# Contributers, funders, acknowledgments, etc.

`[TBD]`

mbjones:
I think the FAQ and other documentation examples are a great idea, and we should strive to have those complete for when the paper comes out, so it can reference them.  So, assigning this to the paper milestone.

cboettig:
[...]

PR #107, opened 2016-09-14
Add faq.md

sdruskat:
This is a draft towards resolving #105.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -72,7 +72,7 @@ Please contact us if you want a new format to be added to CodeMeta.
 
 # CodeMeta for repository managers, publishers, information science
 
-You can make use of CodeMeta's Rosetta Stone quality to figure out suitable standards for your own metadata schema, and make it work with other schemas.
+You can make use of CodeMeta's Rosetta Stone quality to figure out suitable standards for your own metadata schema, and integrate with other schemas and platforms.
 
 # Contributers, funders, acknowledgments, etc.
 


sdruskat:
Contents should definitely be checked by someone with better knowledge of the project...

Rob-Welch:
Here is my feedback on this pull request, based only on my current understanding of CodeMeta (which isn't anywhere near set in stone yet):

- In your 'what codemeta isn't' section, you should stress that while codemeta has a format, it's only a transitional format, and that codemeta is really a specification for conversion between these formats should work, and maybe a test of that implementation (a bit later). So, you could make an application that just reads codemeta files, OR (better) you could make an application that uses an implementation of codemeta (e.g. the codemeta python library) to load in data from any and all metadata formats and serve them to your application in a standard format.

- Similarly, stop talking about the 'codemeta format' and start talking about codemeta standards, libraries etc. We don't want people thinking this is 'just another format'. With the codemeta library, you could read data from many different sources easily into one program to create a searchable repository or index.

- Explain why codemeta incentivises people to release software (because it means they can get citations for the software alone and they don't have to think about writing papers)

- CodeMeta compatible just means that CodeMeta can read in your format. If not, adding CodeMeta compatibility is just a matter of adding to the crosswalk csv file. 

- Uses of codemeta: codemeta for developers of research software would mean 'include some form of codemeta-compatible metadata with your project'. If you want to build an application using codemeta, you want a codemeta library. If you want to build a codemeta library (e.g. you're a C++ user and you want to make a library in C++, not python or ruby) then you want the codemeta spec and the crosswalk.

- 'CodeMeta for repository managers, publishers, information science' - these guys should be implementing the codemeta libraries (or making their own).

mbjones:
Thanks for this!  It's a great start, and I'd like to go ahead and merge it.  But I think we should discuss the "What CodeMeta isn't" section a bit more, and get input from @cboettig and others.  Originally CodeMeta was intended to be mainly the crosswalk, and not be yet another specification.  However, in doing the mapping it became quickly clear that an n:n mapping was not going to work well, and so the CodeMeta concepts list was developed in order to do an n:1 mapping among spaces.  But that makes the codemeta.json actually represent another vocabulary (albeit mostly using existing vocabularies like schema.org).  And as such, it is actually a pretty valuable specification unto itself, and is being used as a standalone format (e.g., within [JORS](http://openresearchsoftware.metajnl.com/)). So, I think as a group we need to revisit this project statement, and maybe indicate that the codemeta.json format is a perfectly viable vocabulary, and has the advantage of a good crosswalk to n other vocabularies.  Thoughts anyone?

cboettig:
I believe this is all covered in https://codemeta.github.io.  If not, issues or PR's against the web source at https://github.com/codemeta/codemeta.github.io to improve docs are more than welcome!!

PR #111, opened 2016-09-15
added initial mappings for cargo.toml (rust package manager)

encima:
I believe `crates.io` adds its own metadata but I have only added tags that pertain to the mainfest structure of a `cargo.toml` file. 

I also could not see a specific key for documentation (I know there is `DocsCoverage`), but that is for a separate discussion and, thus, I have left out the documentation link.


cboettig:
Can't merge multiple PR's to a csv (wouldn't be an issue if the crosswalk was in JSON...)

This conflicts with the just-merged PR #170. Can you take a quick go at resolving that (working from the mobile here so I won't attempt it right now)

cboettig:
p.s. can you also add yourself to the contributors under https://github.com/codemeta/codemeta/blob/master/codemeta.json?

encima:
Will do, I will get this done tomorrow afternoon.

Thanks for the update!
Chris

cboettig:
👍

Issue #127, opened 2017-03-15
From SWH archive- questions about DOAP, SEON, debtags, ADMS.SW, Wikidata, generators and statistics..


moranegg:
First of all let me say that the CodeMeta project is a blessing for understanding
and implementing metadata standards regarding software.

I'm a research intern at [Software Heritage](https://www.softwareheritage.org/) which collects the source code of software projects from a variety of origins. The same source code can be collected from different origins, each providing potentially complementary or conflicting metadata describing the project of which this source code is part, and the people that take part in it.

During the brief research I have conducted over metadata schemas for software projects, there are a
few questions I have in mind for the CodeMeta community:

1. In the crosswalk table the standards bellow are omitted, is there a reason for this omission or just a lack of working hands to add them in? 
  - [x] DOAP
  - [ ] ADMS.SW
  - [ ] debtags 
  - [ ] [SEON ](http://www.se-on.org/ ) (A family of Software Evolution ONtologies)
  - [x] [Wikidata software properties](https://www.wikidata.org/wiki/Wikidata:WikiProject_Informatics/Software#Properties)
   
2. Looking at the ontology, you use schema.org terms, but the schema.org ontology isn't included in the comparison in the crosswalk table. Why ?

3. I was really excited when I found Afron's CodeMeta generator. Are there more tools for
generating or parsing metadata files from one standard to codemeta.json?

4. Do you have statistics about the usage of CodeMeta or any other standards in Github or in general?

5. How can I help?

Thank you for any light you can shed on these dark spots.
Morane

mbjones:
@moranegg Glad its looking useful to you.  Regarding your questions...

- We welcome additional crosswalks.  We've wanted to do DOAP (issue #10) and ADMS.SW (issue #41), but haven't gotten there.  Is debtags the same things as the Debian package format?  If so, that is already in the crosswalk.  Hadn't come across SEON, probably should enter a ticket for that one.  And had thought about Wikidata software, that would be good to add too.
- We've been mapping everything *to* schema.org properties, so we felt it was redundant to include it in the table, but it probably wouldn't hurt.
- His generator is great.  I don't think it produces the most recent version of CodeMeta, so it needs to be patched up for that.  We are working on a generator in R to allow people to upload codemeta to data repositories from R.  Other tooling efforts are definitely welcome.
- Nope, haven't collected any stats.
[...]

PR #29, opened 2016-03-16
Added some concepts to the crosswalk

zoidy:
Added SoftwareContributor, Publisher, Relationship, InteractionMethod,
ProgrammingLanguage. Updated the Software Entities Model accordingly. The other entries already in the table were not updated

See #28 for my justification. InteractionMethod is meant to capture the way the user interacts with the software (service, CLI, API etc)

PR #130, opened 2017-03-31
added DOAP to crosswalk table

moranegg:
Hey,

I'm trying to add the different vocabularies I have asked about in  [#127](https://github.com/codemeta/codemeta/issues/127).

For now, I have only added DOAP after posting an issue on the DOAP project [#29](https://github.com/ewilderj/doap/issues/29), but I didn't add the fact that it could be used with Dublic Core Terms because the DC was already crosswalked.

cheers,
Morane


cboettig:
@moranegg Thanks, this is great, DOAP is now in the crosswalk.  ended up with a manual merge due to shifting stuff around, sorry about that.

PR #143, opened 2017-05-15
Wikidata property mappings

emulatingkat:
Based on my understanding of the Wikidata data properties namespace, these properties are the best matches I could find for the inventory of Codemeta properties. Would these 2 columns of Wikidata mappings be useful for your project?


cboettig:
Looks great, thanks!  It's not clear to me if the id column is useful in this context, so perhaps we might remove that?

emulatingkat:
I included the ID column thinking there could be a use for the URI for the properties.
By adding 'https://www.wikidata.org/wiki/Property:' to the P1072 ID we can get the URI:
https://www.wikidata.org/wiki/Property:P1072

It would be fine with me if they are not included, I'm still learning about and exploring your work. Thanks for the merge!

moranegg:
This mapping is very important for linked data!

Personlly, I prefer having the ID column in the crosswalk table because the
PID are more useful than the actual names when using or querying wikidata.

Le 16 mai 2017 5:19 PM, "Kat Thornton" <notifications@github.com> a écrit :

I included the ID column thinking there could be a use for the URI for the
properties.
By adding 'https://www.wikidata.org/wiki/Property:' to the P1072 ID we can
get the URI:
https://www.wikidata.org/wiki/Property:P1072

It would be fine with me if they are not included, I'm still learning about
and exploring your work. Thanks for the merge!

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/codemeta/codemeta/pull/143#issuecomment-301816464>, or mute
the thread

cboettig:
Thanks for clarifying.  In the other examples I've seen using linked data (eg in the crosswalk table: schema.org, software ontology, doap, Dublin core) has always used the property name in the URI, e.g. name means `schema.org/name` not `schema.org/id1234`.  I see your point about having full URIs, but it is not entirely clear to me how it works in this case.  What would a json-ld or rdf snippet in wikidata context look like?

emulatingkat:
I think the reason for this difference is that Wikidata is multi-lingual. The reason that the labels are not used in the URI is that it would privilege one language over others. Thus the URIs in Wikidata are all numeric.

The RDF for property P1072 is available: https://www.wikidata.org/wiki/Special:EntityData/P1072.rdf

cboettig:
@emulatingkat Ah thanks, that makes sense.  Some of the other ontologies are also internationalized, but I guess they aren't concerned that the root URI shows a linguistic bias, e.g. http://usefulinc.com/ns/doap# .  There's an open discussion about how best to use wikidata aside schema.org / within json-ld here: https://github.com/schemaorg/schemaorg/issues/1186

For the crosswalk however, I might contend that the URIs are not necessary.  Keep in mind that the purpose of the crosswalk is to translate between vocabularies (think `sameAs`) rather than extend them.  So while we absolutely need URIs to write linked data where we want to *extend* our vocabulary with *new* terms from wikidata, it's actually the opposite when we are trying to make the assertion that wikidata `programming language` is the same concept as `schema.org/programmingLangauge`.  A linked data file using full URIs for both could happily maintain these two as completely distinct concepts, (lacking some OWL invocation of sameAs to clarify things).  This makes life harder for the developer working with such data.  

Remember that most columns in the crosswalk are not RDF vocabularies or ontologies, but merely context-specific vocabularies.  So the spirit of the crosswalk is to map these terms to a set of shared terms.  Did that make any sense?

emulatingkat:
@cboettig Thank you for sharing the link to the discussion of wikidata & schema & json-ld. I hadn't seen that yet. I look forward to reading through the full thread.

Yes, your description of the function of the crosswalk makes sense to me.

Issue #149, opened 2017-06-23
spelling differences in Node.js crosswalking and homepage missing


moranegg:
In Node.js/NPM crosswalking the term used is bundledDependencies, I was just looking at some package.json files and discovered bundleDependencies without the d is a popular and discovered that:

> "If this is spelled "bundleDependencies", then that is also honored." (https://docs.npmjs.com/files/package.json)

is this something we want to make note of? by adding both spellings into the crosswalk?

Also, I will do a pull-request to add homepage as url for Node.js.

cboettig:
Thanks, yeah, we could add both spellings (comma separated in the crosswalk cell).  And adding hompage.  A PR for this would be great

moranegg:
oh sorry, i was to quick to PR only the homepage addition.

cboettig:
never too late! if you just push any additional changes to the branch you made the PR on, they become part of the current PR.

moranegg:
great!
thanks for the help PRing :-D

PR #150, opened 2017-06-23
added homepage as url to Node.js in crosswalk table

moranegg:
responding to a part of my #149 issue.
I noticed homepage was missing from the Node.js crosswalking.


cboettig:
Thanks!

PR #158, opened 2017-07-16
link to crosswalk.csv rather than codemeta-crosswalk.md (404)

pdurbin:
When you click "mappings" you get a 404 at https://github.com/codemeta/codemeta/blob/master/codemeta-crosswalk.md 

By the way, I got here because I'm thinking about describing software projects over at https://github.com/IQSS/open-source-at-harvard/issues/3 and feedback there is quite welcome!


cboettig:
Good catch, thanks!

PR #170, opened 2017-10-26
Add Citation File Format mapping

sdruskat:
This PR adds the [Citation File Format](https://github.com/citation-file-format/citation-file-format) mapping to crosswalk.csv.


sdruskat:
Fixes citation-file-format/citation-file-format#21.

cboettig:
Nice.

cboettig:
We should update the web version of the crosswalk table too.  The reading from this table is automatic, but currently we still manually create a placeholder page for the column (which includes some additional background like an icon and an explanation of what the vocabulary being cross-walked does / where it comes from).  See https://github.com/codemeta/codemeta.github.io/tree/hugo/content/crosswalk

cboettig:
p.s. can you also add yourself to the contributors under https://github.com/codemeta/codemeta/blob/master/codemeta.json?

PR #176, opened 2017-12-13
close "You can cite..." sentence logically

katrinleinweber:
Hello!

I noticed that `...cite the CodeMeta schema and project as:` trails off into the next heading. Maybe the quote block got shifted upwards before it was initially committed? Since I [found no commit to revert this disconnect](https://github.com/codemeta/codemeta/blame/fc4310326054b6307b48b24176569ea175fcaad1/README.m), here is a simple fix suggestion that seems logical to me. In case something else should be done to close the sentence, please let me know.

Cheers :-)


cboettig:
LGTM

PR #178, opened 2017-12-20
Update crosswalk table for CFF

sdruskat:
- Reflects CFF-Core version 1.0.2, which focuses on citation metadata only (more software metadata to be included in another CFF module)


cboettig:
Nice, thanks!

Issue #191, opened 2018-09-25
network requirements?


brucellino:
Hello codemeta. :heart:  your work, busy trying to adopt it for infrastructure middleware components. I have a question about requirements in the schema.

There are several terms which address the hardware needs of the item: 
  - `memoryRequirements`
  - `operatingSystem`
  - `processorRequirements`

_etc_ - but I can't see any reference to the networking stack. One of the most important aspects is whether the item is IPv6 ready, whether it supports dual IPv4/IPv6 stacks, _etc_.

Would it be possible to discuss adding this term? 
Thanks!
Bruce

cboettig:
Sounds reasonable to me.  It would be nice to see that term become part of https://schema.org/SoftwareApplication in general as well.  

Currently we have: `memoryRequirements`, `processorRequirements`, `storageRequirements`, `softwareRequirements` ; `networkRequirements` would fit nicely in that list...  Would need some more work though to flush out the relevant vocabulary for specifying these requirements, though maybe a text block would not be a bad starting place.  Thoughts?

brucellino:
Would it be appropriate to send a PR ? if so, where?

PR #194, opened 2018-10-02
Add the term networkRequirements

brucellino:
This adds the term networkRequirements as described in #191

I've checked the various endpoints of the crosswalk and can not find
a case in which this term is expressed, so the tables is blank -
only codemeta will have this term.

It is needed for the proper description of infrastructure products
such as grid and cloud middleware.

The term takes a `Text` value:

  - IPv4 (for IPv4 support only)
  - IPv6 (for IPv6 support only)
  - dualStack (for dual stack support)

### DIFF
--- a/codemeta.jsonld
+++ b/codemeta.jsonld
@@ -66,6 +66,7 @@
         
       "author": { "@id": "schema:author", "@container": "@list" },
       
+      "networkRequirements": { "@id": "codemeta:networkRequirements", "@type": "@id"},
       "softwareSuggestions": { "@id": "codemeta:softwareSuggestions", "@type": "@id"},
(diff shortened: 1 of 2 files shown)


cboettig:
@mbjones thoughts on this?

moranegg:
@brucellino in the modified crosswalk.csv i see that this term isn't  crosswalked with other vocabularies/ontologies. would you know if the term exists in one or more vocabularies?

I would suggest, when adding a term, to verify with all the crosswalked vocabularies the compatible term. This to ensure that the table is up-to-date with the schema.
Or open a task about.

cboettig:
closing as stale

PR #195, opened 2018-10-09
Better example of funder and funding types

tmorrell:
Add funder field to example codemeta file and make it clearer that funding is for specific grant number and title information.  I'm not sure how we'd handle multiple funders and grant numbers - Ideally it would be better to have funding (grant info) be a subset of funder (like DataCite).

### DIFF
--- a/codemeta.json
+++ b/codemeta.json
@@ -153,7 +153,11 @@
   "contIntegration": "https://travis-ci.org/codemeta/codemeta",
   "developmentStatus": "active",
   "downloadUrl": "https://github.com/codemeta/codemeta/archive/2.0.zip",
-  "funder": "National Science Foundation",
+  "funder": { 
+          "@id": "https://doi.org/10.13039/100000001",
+          "@type": "Organization",
+          "name": "National Science Foundation"
+  },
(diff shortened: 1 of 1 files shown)


cboettig:
@tmorrell Thanks, this is nice.  Typically a https://schema.org/funder is expected to be a Person or Organization type, not a text value, e.g. in this case it might be:

```
{
      "@id": "https://doi.org/10.13039/100007523",
      "@type": "Organization",
      "name": "Division of Advanced Cyberinfrastructure",
      "parentOrganization": {
        "@type": "Organization",
        "@id": "https://doi.org/10.13039/100000083",
        "name": "Directorate for Computer and Information Science and Engineering",
        "parentOrganization": {
          "@id": "https://doi.org/10.13039/100000001",
          "@type": "Organization",
          "name": "National Science Foundation"
        }
      }
```

Where we use [FundRef DOIs](https://www.crossref.org/services/funder-registry/) to identify the organization etc.  Possibly that's needlessly complex for an example, maybe

```
"funder": { 
          "@id": "https://doi.org/10.13039/100000001",
          "@type": "Organization",
          "name": "National Science Foundation"
}
```

would be better? 

@mbjones thoughts?

cboettig:
👍

Issue #161, opened 2017-10-11
Should codemeta extend rather than subset schema.org?


cboettig:
Our current ([v2.0](https://raw.githubusercontent.com/codemeta/codemeta/2.0/codemeta.jsonld)) construction of the codemeta context file adopts an explicit subset of terms from schema.org, and adds a 10 extra terms. 

If a user wants to include additional metadata about, say, an associated ScholarlyArticle, grant, etc, this means the user needs to extend the context themselves to include additional schema.org terms:

```
{ "@context": ["https://doi.org/doi:10.5063/schema/codemeta-2.0", "http://schema.org"],
  "@type": "SoftwareSourceCode",
  "citation": {
                   "@type": "ScholarlyArticle",
                   "name": "A paper about this software"
   }
}
```

Otherwise the `"@type": "ScholarlyArticle"` would not be recognized. This case was brought up (without resolution) in #155 , but this same issue will impact including data such as Funding information #160, and in general any extension to codemeta.  Maybe the above solution (let's call that solution A) is just fine, but it feels to me that this discourages rather than encourages the above documentation.  

Additional terms could be added explicitly to the CodeMeta namespace on a case-by-case basis.  That is not unreasonable, but obviously involves considerably more maintenance, approval process, and more releases.  That seems suboptimal (Solution B)

Third solution would be simply to pivot CodeMeta schema so that it is merely an extension, rather than a subset, of schema.org (i.e. schema.org + 10 terms we added that were not in schema.org).  This (Solution C) would, I think, create the least friction in CodeMeta use cases (though being the most permissive, it is potentially more challenging for consumers, though I believe the consumer can merely ignore the additional terms).  To do this, we would simply list our `context` file as `[ {our new terms}, "http://schema.org"]`.    Another benefit (risk?) of this approach is that upstream changes in schema.org terms are automatically reflected in the codemeta context.

npch:
It feels to me that Solution C is conceptually the right approach - particular if we see CodeMeta as sitting in a larger ecosystem of description for research objects and objects related to software.

If this is the case, presumably we would need to get the CodeMeta extensions accepted through schema.org's community process?

cboettig:
@npch Right, I agree that adoption of our additional terms by schema.org would be the long term goal, though it is not necessary for implementing optimization C.  JSON-LD makes it very easy to define your new schema as “schema.org + a few terms” and once we show consistent and significant use of our extension we are better placed to petition schema.org

moranegg:
[...]

PR #201, opened 2018-10-29
Use 'version' instead of 'softwareVersion' in codemeta.json

ProgVal:
'softwareVersion' is only for SoftwareApplication, not for SoftwareSourceCode

### DIFF
--- a/codemeta.json
+++ b/codemeta.json
@@ -163,7 +163,7 @@
     "metadata",
     "software"
   ],
-  "softwareVersion":"2.0",
+  "version":"2.0",
   "dateCreated":"2017-06-05",
   "datePublished":"2017-06-05",
   "programmingLanguage": "JSON-LD"


cboettig:
Thanks, this seems like a reasonable change which would be good for compatibility.

Just a note though: you'll see that CodeMeta is using it's [own context file](https://doi.org/10.5063/schema/codemeta-2.0), which while based mostly on schema.org, is not in fact identical, and that most terms from `schema:SoftwareApplication` are technically valid in the CodeMeta context.  In earlier discussions, our team found it was more useful to express these together, though it's arguable that we should avoid it for greater alignment with schema.org (see #161).  

I do think it is unfortunate that schema.org lacks a clear way to indicate any connection between SoftwareSourceCode and SoftwareApplication; that seems like a natural thing to express in linked data which would make it easier to include both bits of information without breaking from the original type definitions.  (e.g. `softwareRequirements` isn't part of `SoftwareSourceCode` either, but most languages document this in the source code and consider it essential source-code metadata).

band:
I understand the nuance here. But it does seem like trying to specify "all the facets" is an endless activity. Nonetheless more alignment is better.

One question: how do "most languages document 'softwareRequirements' in the source code"? I think I am missing something here. In my experience requirements documents are separate from source code files. And they do not always keep up with the code as it is modified.

cboettig:
@band `softwareRequirements` meaning software dependencies.  i.e. from a in a `package.json` file in npm, a `gemspec` in Ruby, a `DESCRIPTION` file in R, or a `_setup.py` for python/distutils.

PR #202, opened 2018-10-30
crosswalk: add Docker labels

apawlik:
[Stencila](http://stenci.la), as a part of our [dockter project](https://github.com/stencila/dockter), is proposing an addition to the crosswalk for Dockerfile [`LABELs`](https://docs.docker.com/engine/reference/builder/#label) or
(deprecated but still present in some Dockerfiles) [`MAINTAINERs`](https://docs.docker.com/engine/reference/builder/#maintainer-deprecated) instructions.
The LABEL instruction that is un-prefixed or has either the [`org.opencontainers.image`](https://github.com/opencontainers/image-spec/blob/master/annotations.md) prefix,
 or the deprecated [`org.label-schema`](https://github.com/label-schema/label-schema.org) prefix.

The crosswalk uses where possible `schema.org` or `CodeMeta`. However, there are three labels that we have not found a good crosswalk translation:

* ref.name  **org.opencontainers.image.ref.name** Name of the reference for a target (string).
  * SHOULD only be considered valid when on descriptors on `index.json` within [image layout](image-layout.md).
  * Character set of the value SHOULD conform to alphanum of `A-Za-z0-9` and separator set of `-._:@/+`
  * The reference must match the following [grammar](considerations.md#ebnf):
    ```
    ref       ::= component ("/" component)*
    component ::= alphanum (separator alphanum)*
    alphanum  ::= [A-Za-z0-9]+
    separator ::= [-._:@+] | "--"
    ```
* revision  **org.opencontainers.image.revision** Source control revision identifier for the packaged software
* build - used in some images usually to denote the same thing as `revision`


apawlik:
Sure! Will do asap

cboettig:
closing as stale.  feel free to re-open: note that we now use separate files for each crosswalk column, which is also  less likely to result in these merge conflicts as well.

PR #206, opened 2018-12-06
Mapping to ISO 19115-1

tedhabermann:
This crosswalk maps codemeta terms to conceptual paths from the ISO
19115-1 conceptual model. Full XPaths in ISO 19115-3 are included in
Habermann, Mapping ISO 19115-1 geographic metadata standards to
codemeta, PeerJ, in press, and in
https://doi.org/10.6084/m9.figshare.7430282.v1.


cboettig:
awesome, thanks

PR #214, opened 2019-04-23
Provisional DCAT mapping

dr-shorthair:
DCAT uses properties (and classes) from several RDF namespaces.


cboettig:
Nice, this looks good to me!

PR #215, opened 2019-04-26
Wikidata: Add property IDs

larsgw:
Property IDs are the real fields; the labels are just that, labels that 
can change and are available in different languages. On top of that, the 
properties cannot be accessed by label other than via search.

Reverts part of e45d66e

---

Also, adds a bunch of fields.


cboettig:
Thanks, yeah, this is absolutely true, thanks!

Issue #162, opened 2017-10-17
Version and SoftwareVersion


moranegg:
In the CodeMeta terms we can find both, schema:version and schema:softwareVersion.
the first from CreativeWork and the latter from SoftwareApplication.

What's the difference?

Does anyone uses both for different entries?

cboettig:
Yup, they are both inherited from schema.org, so excluding one would just make codemeta unnecessarily incompatible with other schema.org SoftwareApplication descriptions.  Schema.org usually seems to avoid creating this kind of duplication though, so I'm not sure how softwareVersion slipped in (technically `version` is typed to be either numeric or text data, `softwareVersion is only text).  

I always stick with `version`; haven't seen much use of `softwareVersion`.

PR #218, opened 2019-09-26
R Package Description: Proposed changes

clnsmth:
In context of the R Package DESCRIPTION file:
* The "Package" field better aligns the "name" property of codemeta (schema:Thing) than the "Title" field.
* The "Version" field aligns well with the "softwareVersion" property of codemeta.
* The "identifier" property of codemeta (schema:Thing) is not well represented by the "Package" field.

### DIFF
--- a/crosswalks/R Package Description.csv
+++ b/crosswalks/R Package Description.csv
@@ -15,7 +15,7 @@ processorRequirements,
 releaseNotes,
 softwareHelp,
 softwareRequirements,"Depends, SystemRequirements"
-softwareVersion,
+softwareVersion,Version
 storageRequirements,
 supportingData,
 author,[aut] in Author
@@ -43,8 +43,8 @@ isPartOf,
 hasPart,
 position,
 description,Description
-identifier,Package
-name,Title
+identifier,
+name,Package
 sameAs,
 url,URL
 relatedLink,


cboettig:
@clnsmth Thanks Colin, but can you provide more rationale for these statements?  For instance, note that CodeMeta `name` is https://schema.org/name, which, as you will see at that link, is considered as an (OWL) equivalent property to Dublin Core's `title`, http://purl.org/dc/terms/title.  So I think it is pretty clear that DESCRIPTION `Title` field is a `title` in the Dublin Core sense and thus a `name` in schema.org and CodeMeta.  

I think there is also a good case that the DESCRIPTION Package field that contains the package name also operates as an identifier, and not a title.  The package name is clearly not the same thing as the package title (or else the DESCRIPTION field wouldn't have a `Title` field), but it does operate as an identifier that is unique for a given repository (e.g. CRAN).  

I agree entirely that `version` aligns with `softwareVersion`.  Note that it also aligns with `version`, as CodeMeta includes both concepts (one inherited from `CreativeWork` and one inherited from `SoftwareApplication` and probably ought to provide guidance on what to do in this case.

clnsmth:
Thanks for this clarification @cboettig! I'm in full agreement with your rationale.

Should I re-submit this pull request to add the R Package DESCRIPTION `Version` field to the CodeMeta `softwareVersion` property, as well as to provide some guidance on which _version_ to use? Guidance would be to prefer `softwareVersion` over `version` as it's inherited from a more specific type than the generic `CreativeWork`.

cboettig:
Thanks!  see https://github.com/codemeta/codemeta/issues/162#issuecomment-337293008

clnsmth:
Hi @cboettig, the new pull request has been submitted as #224.

PR #219, opened 2019-10-15
Add crosswalk for Leiningen.

ProgVal:
Sources:

* https://github.com/technomancy/leiningen/blob/5ff3e2469050ebc42a9f1f0be5194e39627533be/sample.project.clj
* https://github.com/technomancy/leiningen/blob/5ff3e2469050ebc42a9f1f0be5194e39627533be/leiningen-core/src/leiningen/core/project.clj#L442-L445


cboettig:
nice, thanks!

PR #221, opened 2019-10-15
Add crosswalk for Julia Project.toml file

mkuzak:
Based on https://julialang.github.io/Pkg.jl/v1/toml-files/


cboettig:
nice, thanks! (apologies about travis status, need to fix that).

PR #224, opened 2019-10-21
Update R Package Description.csv

clnsmth:
The "Version" field of the R Package DESCRIPTION aligns with both the "softwareVersion" and "version" concepts of CodeMeta and inherited from "SoftwareApplication" and "CreativeWork", respectively. Generally, "softwareVersion" is preferred over "version" as the former is inherited from the more specific "SoftwareApplication" type as opposed to the more general "CreativeWork".

NOTE: "softwareVersion" is typed text only, whereas "version" is typed to be either numeric or text. Also, the "softwareVersion" concept seems to be less frequently used than "version".

### DIFF
--- a/crosswalks/R Package Description.csv
+++ b/crosswalks/R Package Description.csv
@@ -15,7 +15,7 @@ processorRequirements,
 releaseNotes,
 softwareHelp,
 softwareRequirements,"Depends, SystemRequirements"
-softwareVersion,
+softwareVersion,Version
 storageRequirements,
 supportingData,
 author,[aut] in Author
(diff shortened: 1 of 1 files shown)


progval:
> Also, the "softwareVersion" concept seems to be less frequently used than "version".

Indeed, other crosswalks only have `version`, so we should do the same here for the sake of consistency.

moranegg:
This PR can be merged into V2.1  after modification (in the crosswalk release)
let's add a  2.1 tag to it

progval:
Relevant discussion: https://github.com/codemeta/codemeta/issues/264

Issue #198, opened 2018-10-11
linking between SoftwareSourceCode and SoftwareApplication?


cboettig:
IMHO, schema.org/SoftwareSourceCode and schema.org/SoftwareApplication ought to have a reversible property indicating the relationship between the two.  This would allow us to nest properties which are technically part of SoftwareApplication and not part of SoftwareSourceCode (e.g. `memoryRequirements`) as part of a SoftwareApplication type object, rather than putting them directly as properties of the SoftwareSourceCode object.  

SoftwareSourceCode does have `targetApplication`, but this does not seem to refer to the application that the source-code is defining, but rather a runtime environment (such as an operatingSystem).  SoftwareApplication does not seem to have any property corresponding to identifying the application's source code. 

Thoughts on this?  Shall we open this issue over in https://github.com/schemaorg/schemaorg for discussion?

ProgVal:
> SoftwareSourceCode does have `targetApplication`

Do you mean `targetProduct`?

> but this does not seem to refer to the application that the source-code is defining, but rather a runtime environment (such as an operatingSystem).

The runtime env is referred to by `runtimePlatform`.

`targetProduct` description is vague enough to mean either of this definitions

cboettig:
right, I meant `targetProduct`.  Yes, I agree that `targetProduct`'s definition is hopelessly vague:

> Target Operating System / Product to which the code applies. If applies to several versions, just the product name can be used.

the OS and the Product (application) to built by the code means totally different things to most developers!  (Though perhaps the author of this meant the sense in which a hardware product is often used as a synonym for an OS -- i.e. `"targetProduct": "iphone"` vs `"targetProduct": "iOS"`).  But who knows, like you say, it is pretty vague.  In any event, it would be nice to have an unambiguous property on `SoftwareSourceCode` and a reverse-property on `softwareApplication` connecting these two.

dgarijo:
At the Scientific Software Registry Collaboration Workshop we have also encountered this issue. Many people were confused because SoftwareApplication and SSC are kind of merged in the same object (I know they aren't but many people confuse both). 

In my work, I have used Codemeta in https://w3id.org/okn/o/sd and my own registry, and I had to add the property "hasSourceCode" which links a Software to its SourceCode (inverse of targetProduct, which links a source code to its SoftwareApplication). If something similar is added in codemeta (or suggested in schema.org), then the problem should be solved. 

+1 to add such a property. I will suggest it in schema.org too

[...]

Issue #160, opened 2017-10-11
How should funding or grant metadata be supported in schema.org?


cboettig:
There is no obvious way to identify a grant support.  (see earlier thread #137).  In this issue we would like to promote discussion that could find consensus around proposing an `Award` `@type` or similar mechanism to http://schema.org to better support this.  

@mfenner has suggested the following structure:

```
"funding": {
  "@type": "Award",
  "identifier": "654039",
  "funder": {
    "@type": "Organization",
    "@id": "https://doi.org/10.13039/501100000780",
    "name": "European Commission"
}
```

(Note that CodeMeta already introduces the `funding` property to `SoftwareSourceCode` class, but it is not a schema.org property at this time).

In the above proposal, if I understand correctly, Martin is suggesting that `http://schema.org/funding` would become a new property of `CreativeWork` types (at least of `ScholarlyArticle`, `SoftwareSourceCode`, `Dataset` and other academic products that inherit from `CreativeWork`), and that `Award` would be a new Type/Class that would also inherit from `CreativeWork`.    Note that `funder` is already a property of a `CreativeWork`, so that part would come 'for free' with creation of an `Award` refinement of that type.  

Note that this has some potential for confusion with an existing schema.org property, `award`, http://schema.org/award, which simply takes Text strings as it's data type (and refers to "awards" won by the `CreativeWork`, semantically a rather different idea than the funding support of the project.

Let's hash out some good ideas here and once we have a solid proposal we can open an issue at https://github.com/schemaorg/schemaorg and see what they think.  

cc @sckott @arfon @danielskatz @mbjones @gothub and anyone else who we can get to weigh in here!

cboettig:
@mfenner FWIW, I was also wondering how this would look in terms of expressing the grants section of an academic CV in http://schema.org.  Here is my [off-the-cuff version](http://www.carlboettiger.info/js/vita.json), obviously not valid in today's schema.org.  

```
 {
        "@type": "Grant",
        "name": "The rOpenSci Project",
        "startDate": "2015-11-01",
        "endDate": "2018-10-31",
        "funder": {
          "@id": "https://doi.org/10.13039/100007028",
          "@type": "Organization",
          "name": "Helmsley Charitable Trust"
        },
        "amount": { 
          "@type": "MonetaryAmount", 
          "value": "2900000", 
          "currency":  "USD" 
        }
      },
```

Could clearly be improved, but just for discussion fodder.

mbjones:
[...]

PR #229, opened 2019-11-30
Defining property "hasSourceCode" to link SoftwareApplication and SourceCode

dgarijo:
As we discussed in #198 (and the discussions in the Scientific Software Registry Collaboration Workshop), a link between SoftwareApplication and SoftwareSourceCode is needed for several reasons:
1) Potential adopters from software registries would like to use Codemeta, and their starting point is a software entry (SoftwareApplication) rather than a source code. 
2) Right now codemeta mixes in the spec properties from SoftwareSourceCode and SoftwareApplication. I have seen potential adopters confused on the right use of these properties. Having this link made explicit helps separating SoftwareApplication from SoftwareSourceCode.

In this pull request I have:

- Extended the properties_description.csv with a definition of the relationship proposed.
- Modified codemeta.jsonld with the appropriate domain/range for the property.
- Added an example (in examples/codemeta-software.json) on how to use the property using a SoftwareApplication and Codemeta terms. The example has been validated with https://json-ld.org/playground/

Note: I haven't added an extra line to each code meta mappings (it's the inverse of targetProduct), but will be happy to do so.

### DIFF
--- a/codemeta.jsonld
+++ b/codemeta.jsonld
@@ -75,6 +75,7 @@
       "readme": { "@id":"codemeta:readme", "@type": "@id" },
       "issueTracker": { "@id":"codemeta:issueTracker", "@type": "@id" },
       "referencePublication": { "@id": "codemeta:referencePublication", "@type": "@id"},
-      "maintainer": { "@id": "codemeta:maintainer" }
+      "maintainer": { "@id": "codemeta:maintainer" },
+      "hasSourceCode": { "@id": "schema:SoftwareSourceCode", "@type": "@id"},
   }
 }
--- a/codemeta.jsonld
+++ b/codemeta.jsonld
@@ -63,7 +63,6 @@
       "targetProduct": { "@id": "schema:targetProduct"},
       "url": { "@id": "schema:url", "@type": "@id"},
       "version": { "@id": "schema:version"},
-        
       "author": { "@id": "schema:author", "@container": "@list" },
       
       "softwareSuggestions": { "@id": "codemeta:softwareSuggestions", "@type": "@id"},
@@ -76,6 +75,6 @@
       "issueTracker": { "@id":"codemeta:issueTracker", "@type": "@id" },
       "referencePublication": { "@id": "codemeta:referencePublication", "@type": "@id"},
       "maintainer": { "@id": "codemeta:maintainer" },
-      "hasSourceCode": { "@id": "schema:SoftwareSourceCode", "@type": "@id"},
+      "hasSourceCode": { "@id": "codemeta:hasSourceCode", "@type": "@id"}
   }
 }
--- a/examples/codemeta-software.json
+++ b/examples/codemeta-software.json
@@ -1,4 +1,7 @@
 {
+"@id":"https://w3id.org/okn/i/mint/TOPOFLOW",
+"@type": "http://schema.org/SoftwareApplication",
+"http://schema.org/name": "Topoflow Hydrology model",
 "http://schema.org/author": {
   "@id": "https://w3id.org/okn/i/mint/peckham_scott", 
   "@type": "http://schema.org/Person",
--- a/examples/codemeta-software.json
+++ b/examples/codemeta-software.json
@@ -1,0 +1,24 @@
+{
+"http://schema.org/author": {
+  "@id": "https://w3id.org/okn/i/mint/peckham_scott", 
+  "@type": "http://schema.org/Person",
+  "http://schema.org/name": "Scott Peckham"}, 
+"http://schema.org/citation": "Peckham; S. D. 'Geomorphometry and spatial hydrologic modelling.' Developments in Soil Science 33 (2009): 579-602.", 
+"http://schema.org/contributor": {
+  "@id": "https://w3id.org/okn/i/mint/bolton_bob",
+  "@type": "http://schema.org/Person",
+  "http://schema.org/name": "Bob Bolton"}, 
+"http://schema.org/dateCreated": {
+  "@type": "http://schema.org/Date", 
+  "@value": "2002"}, 
+"http://schema.org/description": "Topoflow is a powerful spatially-distributed hydrologic model for various physical processes in a watershed, with the goal of accurately predicting how various hydrologic variables will evolve in time in response to climatic forcings", 
+"http://schema.org/downloadUrl": "https://csdms.colorado.edu/pub/models/doi-source-code/topoflow-10.1594.IEDA.100174-3.1.0.tar.gz", "http://schema.org/keywords": "snowmelt;precipitation;evapotranspiration;infiltration;channel/overland flow;shallow subsurface flow;flow diversions", 
+"https://codemeta.github.io/terms/referencePublication": [{"@id": "Peckham; S. D. 'Geomorphometry and spatial hydrologic modelling.' Developments in Soil Science 33 (2009): 579-602."}],   
+"https://codemeta.github.io/terms/hasSourceCode": {
+  "@id":"https://w3id.org/okn/i/mint/topoflow_source",
+  "http://schema.org/name": "Topoflow source code",
+  "https://codemeta.github.io/terms/codeRepository":"https://github.com/peckhams/topoflow",
+  "@type": "http://schema.org/SoftwareSourceCode"
+}
+}
+
--- a/properties_description.csv
+++ b/properties_description.csv
@@ -67,11 +67,4 @@ codemeta:SoftwareSourceCode,funding,Text,Funding source (e.g. specific grant)
 codemeta:SoftwareSourceCode,issueTracker,URL,link to software bug reporting or issue tracking system
 codemeta:SoftwareSourceCode,referencePublication,ScholarlyArticle,An academic publication related to the software.
 codemeta:SoftwareSourceCode,readme,URL,link to software Readme file
-,,,
-,,,
-,,,
-,,,
-,,,
-,,,
-,,,
-,,,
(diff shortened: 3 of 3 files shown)


moranegg:
I agree that the usage of both classes might be confusing.
On the other hand, I find also confusing having `codeRepository` and `hasSourceCode` terms. 
Even if  `hasSourceCode` is more generic, I would like having a more informative distinction between the two.

dgarijo:
@moranegg, 
I think it is ok to have both terms, as long as they are well defined and with examples. However, I am open to discussion.

'schema:codeRepository' is directly from schema.org, so I don't think we can change it.
'hasSourceCode' is the property I propose to link a schema:SoftwareApplication to a schema:SoftwareSourceCode object (inverse of schema:targetProduct).

Do you have another proposal that better illustrates the term? I don't mind if we call it differently, as long as there is a term to link schema:SoftwareApplication to schema:SoftwareSourceCode.

alee:
I had a discussion with @dgarijo at the above meeting about whether it made more sense for codemeta to have `SoftwareSourceCode` as the base with a `SoftwareApplication` _targetProduct_ or the inverse, codemeta as a `SoftwareApplication` with a `SoftwareSourceCode` _hasSourceCode_. 

From our perspective as maintainers of a digital repository for source code we lean more towards codemeta as SoftwareSourceCode with a SoftwareApplication targetProduct but obviously it could go both ways..

moranegg:
@dgarijo @alee this is a very interesting discussion !
I looked at the added CodeMeta terms, all are added into the source code class with  `codemeta:SoftwareSourceCode`.

From my understanding, the usage of CodeMeta and the integration of a codemeta.json file in the code, is implicitly saying that CodeMeta is about source code which can be compiled and executed, so it can refer to a software product.

we should try to detail in #198 the use cases for `hasSourceCode` where you don't use `codeRepository`, but I would love to see your use cases, so we can document better the creation of a new property.

We are also working on the Force11's SCIWG CodeMeta task force on aligning the CodeMeta terms into schema.org, so maybe this term should be worked similarly. see [SCIWG issue](https://github.com/force11/force11-sciwg/issues/95) on this.

dgarijo:
@moranegg, @alee thanks for your answers.
I am a little confused, @moranegg. Codemeta is already aligned to schema.org. Why does it need to be realigned?

Codemeta uses codemeta:SoftwareSourceCode, which is an extension of schema:SoftwareSourceCode, to add new properties, and that makes sense to me. What I think is confusing (and my impression is that to other people too) is that codemeta also uses SoftwareApplication properties, which do not describe code, but that distinction is not clear in the spec and examples. For example, things like "memoryRequirements", "operatingSystem", "downloadURL", "installURL", etc describe software applications, not code. This is not a problem of codemeta itself, it's a problem of the documentation.

I think Codemeta should be able to describe both SoftwareApplications and SoftwareSourceCode, I am very happy with its scope. From my point of view, I have a software registry where I describe SoftwareApplications, and I want to point to the SourceCode used for it. It makes sense to use hasSourceCode and codeRepository together, because hasSourceCode points to the identifier of the code and describes metadata such as the language it's written in, while codeRepository points to the code repository where the code lives (github, or my server). I can only think of a case where you would use hasSourceCode but not codeRepository, which is when you want to state that an application is written in a programming language, but you don't have the pointer to the source code.

tmorrell:
The realignment is to see if any of the extension terms can fit under newer releases of schema.org. For example, if we can replace codemeta:SoftwareSourceCode,funding with schema:SoftwareSourceCode,funding as in #160 we'd make codemeta even more sustainable.

Adding codemeta:SoftwareApplication,hasSourceCode makes a lot of sense to me, as do the documentation changes. We can then try to get it adopted by schema.org.

dgarijo:
Ah, I understand now. Thanks!
Please let me know if you need some other changes to this PR.

cboettig:
@mbjones Any objections to merging?  It looks good to me, though of course, if we adopt the proposal that codemeta inherits/extends schema.org instead of subsetting it, we get this for free...

https://github.com/codemeta/codemeta/pull/229/files#diff-677f920ff0ca91c4f00013b699382a47R79

As this change impacts the JSON-LD context file, we will want to think about plans for cutting a new release and a new DOI (though that might await some possible further changes, e.g. re `schema:Grant` and other terms in schema.org that might also be declared explicitly into codemeta).

codemeta.jsonld
```diff
@@ -75,6 +75,7 @@
       "readme": { "@id":"codemeta:readme", "@type": "@id" },
       "issueTracker": { "@id":"codemeta:issueTracker", "@type": "@id" },
       "referencePublication": { "@id": "codemeta:referencePublication", "@type": "@id"},
-      "maintainer": { "@id": "codemeta:maintainer" }
+      "maintainer": { "@id": "codemeta:maintainer" },
+      "hasSourceCode": { "@id": "schema:SoftwareSourceCode", "@type": "@id"},
```
mbjones: Final comma on this line is a syntax error and needs to be removed.

Also, it would be better if `hasSourceCode` were grouped with the other schema.org properties following the patterns in the file so far.

mbjones: You can see the error on the JSON-LD playground:  https://json-ld.org/playground/#startTab=tab-expanded&json-ld=https%3A%2F%2Fraw.githubusercontent.com%2Fcodemeta%2Fcodemeta%2F61f2527b69329e6bf1fae28e6cbb9340b8e554cc%2Fcodemeta.jsonld

mbjones:
I noted a syntax error above. Other than fixing that up, I think the general idea of `hasSourceCode` and the specific implementation and definition are all great.  So, let's get the context file to validate and then we can merge.

codemeta.jsonld
dgarijo: I changed it, there was a typo. It should be fixed now. Sorry!

dgarijo:
I validated it: https://json-ld.org/playground/#startTab=tab-expanded&json-ld=https%3A%2F%2Fraw.githubusercontent.com%2Fcodemeta%2Fcodemeta%2Fc6bf60ca58418b7e8fe56b68165b287aebf64822%2Fcodemeta.jsonld
I wrongly added the "schema" prefix, but since it's a new property it should be from codemeta. Sorry about that.

mbjones:
@cboettig Given that this is a schema change, we should probably merge into a branch for the time being as the master branch is used for validation in the tests.  So, if we create a branch for the next schema release version and merge this PR there, all should be ok until we are ready to cut a new release.

dgarijo:
I agree with that. The master branch is used in many projects just as something to be synchronized iwth the latest release. You could have a "develop" branch and suggest contributions there for testing.

mbjones:
Sounds good.  Looking into why the tests in Travis are failing, I think the PR has a mismatch between some of the documentation files and the new changes in the jsonld file.  I fixed the test issues on master, so now it looks like just a problem in the PR with the new property not matching.  You can see the error by running `python3 scripts/aggregate.py` from the root of the branch directory, which gives the error: ```Error in codemeta-V1.csv: property names hasSourceCode and  should be the same```  @cboettig is the issue obvious to you?

dgarijo:
please let me know if I need to change anything else. Thanks!

cboettig:
@mbjones yeah, apologies about the travis checks.  I think should be resolved in https://github.com/codemeta/codemeta/pull/231.

dgarijo:
@cboettig @mbjones 
Any updates on this PR?

codemeta.jsonld
moranegg: > Also, it would be better if `hasSourceCode` were grouped with the other schema.org properties following the patterns in the file so far.
@mbjones the `hasSourceCode` is not a schema.org property even if it makes sense as a schema.org property.

moranegg:
@dgarijo I think this PR is waiting for a CodeMeta v3 release ;-)

mbjones:
@moranegg The tests are still failing on the PR, and so we can't merge until those are passing. Plus, the PR seems to have conflicts that need to get resolved as well. Once those are taken care of, we can merge, and then include it in a 3.0 release.

dgarijo:
@mbjones by "taken care off" does that mean me?
When I submitted the PR, there were no merge conflicts. The Travis build was failing due to an issue that was not obvious, and that was claimed to be fixed in #231.

mbjones:
@dgarijo That's up to you, but would be appreciated. Eventually either Carl or I would get to it in the process of preparing a 3.0 release, but I am fully swamped for the next several weeks or more and don't anticipate having any cycles to spend on this. So I was just pointing out that those were the barriers to merging this particular PR.

moranegg:
This PR is a candidate for discussion, it changes the spec, we should label it #discussion.
@dgarijo: there is a fix to be made that @progval is commenting directly on this PR

This issue will be open for discussion until March 15th and we will start a vote from March 15th until March 25th 2023

progval:
Any reason it shouldn't be this?

```
"hasSourceCode": { "@id": "codemeta:hasSourceCode", "@type": "schema:SoftwareSourceCode"}
```

This allows linking with the `SoftwareSourceCode` object, in addition to allowing an URI to the object; similarly to other properties like `author`.

dgarijo:
That makes sense to me. Do you want me to edit the PR?

progval:
Sure, thanks!

Also, could you rebase on the current `develop` branch? There are some merge conflicts

moranegg:
This PR will be deleted when #300 will be integrated into `develop` branch

progval:
Closing in favor of #300 (which resolves merge conflicts, and adds the reverse property `isSourceCodeOf`)

kavanpatel7020 (commented):
dfrg

PR #230, opened 2019-12-06
removed duplicate key 'version'

jspaaks:
looks like the rename from https://github.com/codemeta/codemeta/pull/201/ accidentally created a duplicate. The other ``version`` is at line 10:

https://github.com/codemeta/codemeta/blob/18b248d9cf492d078ef6d2dde115e1aa137bbcdb/codemeta.json#L10

### DIFF
--- a/codemeta.json
+++ b/codemeta.json
@@ -163,7 +163,6 @@
     "metadata",
     "software"
   ],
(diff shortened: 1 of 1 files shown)


cboettig:
👍 thanks!

PR #235, opened 2020-03-05
Clean up usage of author/creator in favor of just using author

amoeba:
As discussed in #223, codemeta has so far used "author" and "creator" both inconsistently and interchangeably. As decided there, I've gone through the repo and removed the "creator" term from the context file and updated relevant files appropriately. Any crosswalks or other files referencing "creator" had whwatever their mapping was moved to "author", if appropriate.


cboettig:
thanks, LGTM!  @mbjones okay to merge?

mbjones:
LGTM, 👍 on the merge. 

How are we going to handle the next release that this would be in?  It seems we have a few things to clean up for a release, and we need to decide if this would be 2.1 or 3.0, depending on its compatibility.  Are we ready for such a thing?

cboettig:
@mbjones excellent point, yeah we should have that discussion.  Opened a separate thread #236

PR #239, opened 2020-03-20
Set type of 'affiliation' to Organization.

progval:
To match codemeta.jsonld (which aliases it to http://schema.org/affiliation, which is defined as Organization).
Syncs with a website change: https://github.com/codemeta/codemeta.github.io/commit/6ccca7d90bb5a2d4d39ebc5a356edd5301e8bef4


moranegg:
This PR looks good! this is a bug fix, no major change.
if no more comments are added we will merge this PR at the end of March 2023

PR #248, opened 2020-09-28
Change context to https

tmorrell:
Schema.org is now recommending the use of https for context linking https://github.com/schemaorg/schemaorg/issues/2578#issuecomment-698759677. They have recently changed the configuration of schema.org and http:// redirection no longer works. 

Why make this change? The validation example in https://docs.ropensci.org/codemetar/articles/articles/validation-in-json-ld.html no longer works; you'll get an error 'loading remote context failed", url = "http://schema.org"'. Changing the codemeta url to a version that uses https (e.g. https://raw.githubusercontent.com/caltechlibrary/convert_codemeta/master/codemeta.jsonld) clears the error. I've run into the same issue in Python.

There will probably need to be a new release and a new DOI, since the current DOI is tied to the 2.0 release.

### DIFF
--- a/codemeta.jsonld
+++ b/codemeta.jsonld
@@ -2,7 +2,7 @@
   "@context": {
       "type": "@type",
       "id": "@id",
-      "schema":"http://schema.org/",
+      "schema":"https://schema.org/",
       "codemeta": "https://codemeta.github.io/terms/",
       "Organization": {"@id": "schema:Organization"},
       "Person": {"@id": "schema:Person"},


mbjones:
Thanks, @tmorrell this has indeed been raised across schema.org, and we've been discussing it at https://github.com/ESIPFed/science-on-schema.org/issues/52 That long thread with many offshoots and references to other conversations clearly concluded that https://schema.org/ was preferred at this point in time (despite starting at a different assertion). I agree codemeta should make the same change, and I agree it will need to be incorporated in the next release, so I will tag it as such.

Anyone who has feedback on this after reading the thread I linked above carefully, as it covers many issues that would like arise here, please chime in.

moranegg:

moranegg:
Since no other comments have been added and last comment was about keeping `http`. I'm removing the v3.0 tag from this PR.

Issue #232, opened 2020-01-15
Write proposal for the schema.org adoption of the CodeMeta terms


moranegg:
On Force11's Software Citation Implementation Working Group we have started working on a proposal for the schema.org adoption of the CodeMeta terms:

  - [ ] softwareSuggestions
  - [ ] maintainer
  - [ ] contIntegration
  - [ ] buildInstructions
  - [ ] developmentStatus
  - [ ] embargoDate
  - [ ] funding
  - [ ] issueTracker
  - [ ] referencePublication
  - [ ] readme
  - [ ] hasSourceCode (not a CodeMeta term, but was recently discussed here on https://github.com/codemeta/codemeta/issues/198)

We will post here our proposal for each term or all terms for community feedback and then post it on the schema.org repository.

You are welcome to see and contribute comments to our working documents:
- Workflow: https://docs.google.com/document/d/1s2mUS9JjIOq-4-0rZTFNaTO7W9umNF4mCf-uypPwGsQ/edit?usp=sharing
- Terms dashboard: https://docs.google.com/spreadsheets/d/1VmUBdgaC7mtj7gDNqTBzPUdosux1R26Cyp7qA7WHkNA/edit?usp=sharing

dbouquin:
Proposal text for softwareSuggestions:

It is sometimes useful to specify when a software package is helpful, but not required, to run a specific code. For instance, something like the Python module "[Palettable](https://jiffyclub.github.io/palettable/)" could be used to visualize results and make a graph more accessible to people with visual impairments, but it may not be required to render a graph.

A new property named "optionalDependencies" could be used to make distinctions between true dependencies that are required to run software, and optional dependencies that are just recommended. A property like this could be similar to Wikidata's “[software add-on (Q70383937)](https://www.wikidata.org/wiki/Q70383937).”

agbeltran:
@dbouquin Just to add the links to track back the reasons behind the 'softwareSuggestions' property as I could find:
- here there is an example of codemeta using 'softwareSuggestions': https://github.com/codemeta/codemeta/blob/7243575791f487106dd879e9c10974449e00949b/examples/codemeta.json#L43
- here is the DESCRIPTION file for the R package being described: https://github.com/ropensci/codemetar/blob/master/DESCRIPTION
[...]

PR #249, opened 2020-10-28
Switch maintainer to schema.org version

tmorrell:
Switch maintainer from a codemeta term to the pending schema.org term (https://schema.org/maintainer), as discussed by the Force11 working group in https://github.com/codemeta/codemeta/issues/232. I think I found all the instances, but let me know if I missed any.

This should be tagged for the 3.0 release.


mbjones:
Thanks, @tmorrell Please retarget this PR to the `develop` branch, rather than `master`.

PR #251, opened 2020-11-25
Update crosswalk table

tmorrell:
This PR adds bio.tools to the main crosswalk table (It was previously added to the crosswalk folder, but never made it the main table). I modified the split.py script so it assumes you're running it in the scripts directory (like aggregate.py). I also enhanced the Zenodo mapping.


cboettig:
LGTM. @mbjones ?

mbjones:
Overall looks good, but I'm curious why the path had to change on `split.py` to make it relative (and which hard codes UNIX-style path separators now). Was this run from a different directory before? Will it break any CI pipelines? If that looks good to you @cboettig then I am good with merging it into develop. I added it to the v3.0 release milestone.

tmorrell:
Apologies for the delay in getting back to you! I've updated `split.py` so it uses the same file definitions as `aggregate.py`, which should avoid the relative path issues.

moranegg:
Can you divide into 2 separate PR?
PR 251.1 for the crosswalk additional terms (with the `aggregate.py` changes) 
PR 251.2  for the changes in `split.py` .

Thanks!

PR 251.1 issue doesn't changes the spec and should be resolved for V2.1. label #crosswalk
PR 251.2 if no more comments are added we will merge this PR at the end of March 2023 (included into v3.0)

progval:
>  PR 251.2 if no more comments are added we will merge this PR at the end of March 2023 (included into v3.0)

that's a refactoring, no need to wait for v3.0 to merge it

progval:
Superseded by https://github.com/codemeta/codemeta/pull/291 and https://github.com/codemeta/codemeta/pull/292 (which don't conflict, as they don't edit the main crosswalk)

PR #257, opened 2021-06-30
remove softwareSuggestions

tmorrell:
remove the softwareSuggestions per discussion in https://github.com/codemeta/codemeta/issues/233


moranegg:

Many +1 votes are already noted on the issue #233

moranegg:
In any case this PR should be tagged v3.0

moranegg:
Rereading the issue, this PR is implementing the removal but I'm not sure that the suggestion was just to remove a term:
copied from the issue: 
- Proposed: addRecommendations: Additional software to enable optional features, code development, interoperability, etc

Since this is a `breaking` change, I would like to reopen the discussion for v4.0, as the community of infrastructures that might be using it, can be affected.

Issue #243, opened 2020-04-29
Link to continuous integration service


agbeltran:
Codemeta currently has a ```contIntegration``` term with type URL and definition ```link to continuous integration service```.

This issue proposes to expand the label to ```continousIntegration``` and consider changing the definition to ```Link to the instance of the continuous integration service for the software code being described.``` The type would remain as URL, and an example value is https://travis-ci.org/github/codemeta/codemeta

This doesn't change the previous intention but provides more clarifications on what should be included. There is no property to indicate what is the continuous integration service being used (Travis, Jenkins, etc) and that would need to be deducted by the user (but we are not planning to provide this as a machine readable element unless there is a strong requirement to do so in another property). 

No other vocabularies seem to include this. See further analysis at: https://docs.google.com/document/d/1R1zFMYY9uRUzLkYS70m5u63W1Ig-iJ1L6oVVwyY5p6g/edit

moranegg:
I agree with having the longer term `continuousIntegration`.

I would suggest the following definition with an example:
> Link to the instance of the continuous integration service for automation workflows (e.g Travis, Jenkins, etc.)

arfon:
:+1: to the more readable `continuousIntegration`.

dgarijo:
+1 to `continuousIntegration` for consistency (e.g., `developmentStatus` is not `devStatus`)

mfenner:
+1 by the Force11 Codemeta Task Force regarding `continuousIntegration`.

tmorrell:
Implemented in PR https://github.com/codemeta/codemeta/pull/258

moranegg:
Further discussion will be dine on #258

progval:
Merged! (through #302 , which is #258 with merge conflicts resolved)

PR #258, opened 2021-06-30
change contIntegration to continuousIntegration

tmorrell:
per discussion in https://github.com/codemeta/codemeta/issues/243


progval:
In order to minimize disruption of v3, would it make sense to keep `contIntegration` as a (deprecated) alias of `continuousIntegration`? Like this:

```
"continuousIntegration": { "@id": "codemeta:continuousIntegration", "@type": "@id"},
"contIntegration": { "@id": "codemeta:continuousIntegration", "@type": "@id"},
```

moranegg:

dgarijo:
@progval you bring up an interesting point. 
I think having the alias will confuse people (e.g., I am new to codemeta and I see 2 properties, which one to use?). I would only support it if we add a status next to it, something like:

`"continuousIntegration": { "@id": "codemeta:continuousIntegration", "@type": "@id", "status":"deprecated"},`

But then we have to define somewhere what this `status` means. Maybe it's just better to do the change and document it in the release.

progval:
I'm not sure that's allowed in JSON-LD anyway. And sadly, no comments in JSON :(

> Maybe it's just better to do the change and document it in the release.

Yeah, that's less confusion

proycon:
Though I agree with the namechange, I'm also a bit conflicted on breaking backward compatibility.

Maybe we could publish an additional JSON-LD file that people can include in their context and that covers 2-to-3 migration, such as mapping `contIntegration` to `continuousIntegration`?

moranegg:
Since we are striving with the v3.0 to keep it compatible with the v2.0, we will deprecate the term `contIntegration` while keeping it in the jsonld and adding the mention deprecated in the properties description.
With that we will add the new naming as an additional term: `continuousIntegration`.

mbjones:
The whole reason we were thinking of naming this new version 3.0 is there were proposals for incompatible changes. In general we have been following semantic versioning, so if the decision is made to stay compatible, then I think we shou;d keep the version number in the 2.x series.

mbjones:
Oh, and if we do, then we need to inspect other PRs to be sure there are no breaking changes.

moranegg:
The plan with v3.0 is add additional properties while keeping the old name
with a "deprecated" comment in the properties description.
Which means that v3.0 while stay backward compatible but still changes the
vocabulary with new properties.
This is why I believe we can move to a v3.0 naming scheme.
In the v4.0 we will delete deprecated terms.

mbjones:
Again, deprecation alone does not warrant a major version bump under semantic versioning. New feature additions alone are indicated by a minor version change. Details follow from https://semver.org/

Given a version number MAJOR.MINOR.PATCH, increment the:

- MAJOR version when you make incompatible API changes
- MINOR version when you add functionality in a backward compatible manner
- PATCH version when you make backward compatible bug fixes

So, I still think this should be version 2.2.0 if we have backwards compatibility. And I find semantic versioning helpful when understanding upgrade paths. Deviating from it causes unnecessary confusion, as many people will think a major version bump is incompatible and won't upgrade as soon.

moranegg:
Hi Matt,

I do agree that this should be clear when we release the next version.
We opted in the PRs for an in between solution with the deprecation and we can rename it v2.2.
On the other hand there is more social impact when releasing a MAJOR version and it's been a while we are promising a v3.0.

From my point of view there are two options with this upcoming release:
1. naming the release 2.2 and labeling all issues and PRs 2.2. Then  we can quickly move to a v3.0 release. 
   -  the drawback here:  Having v2.2 and v3.0 released closely, might create more confusion about which release should be used.
2. changing the PRs to delete the duplicated terms and releasing a v3.0
   -  the drawback here: Not providing backward compatibility with the next release, it's v2.1 or v3.0 (technically it means that you can't use for example a new v3.0 property `hasSourceCode` or adding natively a role to an author while at the same time using an old v2.1 property with the old name (e.g `embargoDate` and `contIntegration`).

I would like to start a vote on that and give until Monday June 19th.
For that I'm starting a poll in the Discussions:
https://github.com/codemeta/codemeta/discussions/306

Thanks for the valuable feedback.

progval:
Replaced by https://github.com/codemeta/codemeta/pull/302

Issue #236, opened 2020-03-05
Next release?


cboettig:
Are we ready for a new release of the the [codemeta.jsonld](https://github.com/codemeta/codemeta/blob/master/codemeta.jsonld) context file?  Do we do a major or a minor release? (migrated from https://github.com/codemeta/codemeta/pull/235#issuecomment-595388249)

mbjones:
Yeah, I was examining our CodeMeta files in the Google structured data testing tool, and it didn't recognize it as schema.org because wrap the SO terms. So I think we should reconsider that, although I don't seem to be able to replicate it atm.

We also need to ensure that our namespace ends with a `/`, as it currently means our terms append onto the namespace without a separator.

I do like an identifier that resolves to exactly the bytes of our context file, as it makes it self-referential.  That said, if you'd prefer, we can use a namespace that uses a regular https address and redirect from there, which is what we do with our other schemas.  It certainly wasn't easy to get the DOI redirect to work right with the DOI redirection service.

moranegg:
## open issues that might be suited for V3 evolution
- add `review` or `reviewedBy` : https://github.com/codemeta/codemeta/issues/177
- API entry point: https://github.com/codemeta/codemeta/issues/183
- mailing list or `contactPoint` :https://github.com/codemeta/codemeta/issues/166
- link fromSoftwareApplication to SSC with `hasSourceCode`: https://github.com/codemeta/codemeta/issues/198
- input/output formats: https://github.com/codemeta/codemeta/issues/188

moranegg:
When we are ready to move to the CodeMeta v3 or v2.1 we need to add a crosswalk between `v2 -> v3`, like the already existing crosswalk from `v1 -> v2`

mbjones:
[...]

Issue #216, opened 2019-06-29
Codemeta content negotiation is not correct


dgarijo:
Hello,
I want to use Codemeta on our next version of OntoSoft. However, the content negotiation of the schema is not correct. When I do 
`curl -sH "accept:application/json+ld" -L https://codemeta.github.io/terms/`
I am expected to obtain https://raw.githubusercontent.com/codemeta/codemeta/1.0/codemeta.jsonld, i.e., a machine readable description of the schema. Instead, I am redirected to the human-readable html page.
I don't think you can enable content negotiation on GitHub pages, but maybe having a w3id would be desirable.

mbjones:
@dgarijo you make a good point. We published the 2.0 version using the doi https://doi.org/10.5063/schema/codemeta-2.0, and our intention was that would be the context for the terms.  So this resolves properly:

curl -sH "accept:application/json+ld" -L https://doi.org/10.5063/schema/codemeta-2.0

We meant to switch the namespace to that, but it seems we didn't and we are now going to have the github.io URL as the context URI, which could cause issues in the long run for us.  So, @cboettig, what do you think of a new release using the doi as the context?  Alternatively, we live with it at the current github.io URL, and accept that we can't redirect it to the schema, and that we may not be able to maintain it over the long term depending on what happens with github.io.  Other options?

dgarijo:
Some suggestions based on my experience:

I think that a raw.githubusercontent URL may not be the best for production. It would be better to give it a more stable URI, even if you use a github page. Or even better, you tag a release and have the json representation there, so you have a permanent link such as https://github.com/codemeta/codemeta/archive/2.0.jsonld

Do you have control on the content negotiation for https://doi.org/10.5063/schema/codemeta-2.0? If so, in the htaccess you may add a 303 redirect to the github.io html documentation or the jsonld file depending on the header request. Right now doing: 

curl -sH "accept:text/html" -L https://doi.org/10.5063/schema/codemeta-2.0

also returns the json ld file, instead of https://codemeta.github.io/terms/

mbjones:
Thanks @dgarijo Personally I much prefer to keep the version in the namespace, as it makes it much clearer which version of a vocabulary people are using in particular instance documents. But I agree we have some namespace cleanup to do here.  This partly depends on the recent changes at DataCite on how much control we have over content negotiation for the json-ld document -- I need to look into that further.

dgarijo:
All right, please keep me posted :)
[...]

PR #274, opened 2022-03-22
Switch funding to schema.org 

tmorrell:
Switch funding from a CodeMeta term to the schema.org term (https://schema.org/funding), as discussed in https://github.com/codemeta/codemeta/issues/160. I changed the example codemeta.json file to put funder inside of funding, but this is not enforced by the schema (you can still have a top-level funder field).

This should be tagged for the 3.0 release.


megasanjay:
Hi,

I'm just checking in to see what the status of this PR is. I noticed that the `funder` object will now be assimilated with the `funding` object. Does this also mean that multiple funders would be supported with this syntax?

tmorrell:
You have the option to include 'funder' in the 'funding' object, but it's not required. I think it makes sense to include funder under funding because they are usually directly linked.

Feel free to test out the codemeta.jsonld file in this PR and leave a review with how it works. Once there is some feedback it can be merged into the develop branch for the eventual v3 release https://github.com/codemeta/codemeta/issues/236

megasanjay:
Awesome. We will update our application in preparation for this. The funder array makes more sense anyway but we shall wait for the v3 release. In the meantime, I'll work on a PR for the update on the code meta generator.

moranegg:

Also to be tagged v3.0 release if accepted by vote.

stain:
Ping to merge this -- switching `funding` is needed as it's been live in schema.org since 14.0 (2022-03).

The comments says "This should be tagged for the 3.0 release" however it seems the 3.0 release is already tagged and listed in https://github.com/codemeta/codemeta/releases 

However 3.0 is not listed in  https://codemeta.github.io/jsonld/ and https://github.com/codemeta/codemeta/#codemeta-schema or our own codemeta (#299) so perhaps 3.0 can be considered a "quiet" summer release candidate -- perhaps we can consider 3.1 to have `funding` fixed as well as the namespace mapping to w3id as @dgarijo says in #321 #216.

I'm adding the remaining terms to RO-Crate 1.2 but we've always had `funding` with the `http://schema.org/` namespace as it was used in their examples before I helped push it live.

progval:
Oops, that was an oversight. I guess this will have to wait for v4.0

dgarijo:
A reminder @moranegg @progval to include this in the next revision of CodeMeta, since we have https://schema.org/funding

dgarijo:
The PR for updating the develop branch: https://github.com/codemeta/codemeta/pull/470 (blocking progress here)

Issue #272, opened 2022-02-14
Add additional container values


tmorrell:
The author term has a `@container: "@list"` value, which indicates it should be an ordered array. There are other codemeta terms that should probably have container values, which will make codemeta more reproducible (it ensures there will always be an array, even if there is only one object):

**contributor**
Has same use case as author, and is shown as an array in https://github.com/codemeta/codemeta/blob/master/codemeta.json.

**keywords**
Is shown as an array in https://github.com/codemeta/codemeta/blob/master/codemeta.json; could use `@set` to indicate the order is not important

**editor, copyrightHolder, funder**
These all have person or organization type, and it seems likely that there could be one or more per piece of software.

If folks think this sounds reasonable, I can put together a PR.

cboettig:
The intent of declaring authors as an array is, as you note, that order matters, since somehow we academics have deeply ingrained that notion with respect to authorship.  I'm not sure it was a good idea though.  It is a principle of RDF (and of good data management elsewhere, including relational DBs right?) that data relationships should be _semantically explicit_, not hinted at in structural ways  such as 'order'.  

On a more immediately practical note, I worry about the array declaration not being as compatible with vanilla schema.org use, including prevailing practices for dealing with academic authorship order in schema.org; but maybe @mbjones, @mfenner or others would know better here.

mbjones:
We use the `@list` structure on science on schema.org as well, and have a [writeup describing the uses for `@list`](https://github.com/ESIPFed/science-on-schema.org/blob/develop/guides/GETTING-STARTED.md#json-ld-list) in the upcoming release that may be helpful. I think adding the container to clearly ordered fields makes a lot of sense, and would prevent complications for people that aren't familiar with the unordered nature of JSON-LD.

tmorrell:
I've made an initial PR for contributor and keywords, which maps to the behavior of https://codemeta.github.io/codemeta-generator/ in including an array for all cases. 

I think **funding**, **funder**, **editor**, and  **copyrightHolder** are other candidates for containers, but these would be a larger change in behavior.

PR #275, opened 2022-03-22
Add containers for contributors and keywords

tmorrell:
As discussed in https://github.com/codemeta/codemeta/issues/272


moranegg:

PR #280, opened 2022-08-18
Remove type restriction of "releaseNotes"

progval:
https://codemeta.github.io/terms/ documents it as "Text or URL" (like schema.org), but
`"@type": "@id"` means only URLs are allowed.


tmorrell (approved):
Makes sense to me. Should be included in the v3 release, whenever that gets organized

moranegg:
This PR looks good but changes the spec (changes the context)
if no more comments are added we will merge this PR at the end of March 2023 (included into v3.0)

PR #285, opened 2022-11-18
replace date_published with publication_date in zenodo crosswalk

vuillaut:
`date_published` is not a Zenodo keyword but `publication_date` is

see https://developers.zenodo.org/#representation


vuillaut:
However, it is probably not the right date either in the crosswalk, if I believe the terms definition:
```
dateCreated | Date or DateTime | The date on which the CreativeWork was created or the item was added to a DataFeed.
dateModified | Date or DateTime | The date on which the CreativeWork was most recently modified or when the item’s entry was modified within a DataFeed.
datePublished | Date | Date of first broadcast/publication.
```

Then, it is `dateModified` that should be used in the crosswalk to `publication_date`

tmorrell:
I disagree with the proposed mapping to dateModified. `publication_date` in Zenodo is exactly analogous to `datePublished`, since it's set by the user for whenever they believe the work was published. 

`dateModified` would map to the Zenodo `updated` field, which is a system field instead of a metadata field. I'm honestly not sure how we'd indicate that in the crosswalk file. 

If we're saying the crosswalk should be a direct mapping to the api endpoint, then all the zenodo terms should prefixed with metadata (e.g. metadata.publication_date). But that's more annoying for people reading it.

vuillaut:
> I disagree with the proposed mapping to dateModified. `publication_date` in Zenodo is exactly analogous to `datePublished`, since it's set by the user for whenever they believe the work was published.
> 
> `dateModified` would map to the Zenodo `updated` field, which is a system field instead of a metadata field. I'm honestly not sure how we'd indicate that in the crosswalk file.
> 
> If we're saying the crosswalk should be a direct mapping to the api endpoint, then all the zenodo terms should prefixed with metadata (e.g. metadata.publication_date). But that's more annoying for people reading it.

I certainly misunderstand the terms' definition, then.
I made the modification according to  @tmorrell's comment.

Issue #246, opened 2020-06-29
Proposal for the adding the `hasSourceCode` property in schema.org and CodeMeta


moranegg:
This issue is part of the SCIWG CodeMeta task force aiming to add all additional properties into schema.org.

Text proposal for the adding the `hasSourceCode` property in schema.org and CodeMeta

It can be useful to unambiguously define the relationship between a softwareApplication and its softwareSourceCode. Adding an element to act as the inverse to the element “targetProduct” (Target Operating System / Product to which the code applies) would achieve this goal. 

For example, unpaywall is a softwareApplication. One product to which unpaywall applies (its “targetProduct”) is the browser, Chrome. Put more simply, unpaywall is for Chrome. To define an inverse, it would be useful to designate where the softwareApplication is from. If a softwareApplication is like unpaywall, its source code is available on a public repository that could be identified using schema.org’s term “codeRepository.” If unpaywall’s source code was instead available in the appendix of an article, using the term “citation” (a reference to another creative work) would be ambiguous as it would not be possible to distinguish a “citation” to the source code appendix from a “citation” to a website or other publication about the code. For this reason we propose the addition of the term “hasSourceCode” to schema.org. 

Furthermore there is the possibility to reference the archived software source code in Software Heritage  using the SWHID and giving a direct access to the artifact itself, for example the Darktable repository snapshot (4 May 2017, GitHub) :swh:1:snp:c7c108084bc0bf3d81436bf980b46e98bd338453

The utility offered by “hasSourceCode” is similar to the utility offered by the terms “isPartOf” and “hasPart”— the former allows you to specify that a specific creative work is part of another creative work, while the latter is its inverse. “targetProduct” allows you to specify that a creative work applies to another creative work, while “hasSourceCode” allows you to specify that a creative work is executed from another creative work.

Note: following https://github.com/codemeta/codemeta/issues/198

* This proposal was written in collaboration with @dbouquin

dgarijo:
+1 to add this property, which I also suggested in the pull request: https://github.com/codemeta/codemeta/pull/229 (with an example, if I remember correctly)

mfenner:
The Force11 Codemeta task Force supports adding this property.

rebeccaringuette:
@moranegg likely 4.0 topic

tmorrell:
This issue is for making a proposal to schema.org for the terms "hasSourceCode" and "isSourceCodeOf", which were added to CodeMeta in v3.0. So it's not directly related to the 4.0 release.

PR #300, opened 2023-05-25
Define properties `hasSourceCode`/`isSourceCodeOf` to link SoftwareApplication and SourceCode

progval:
This is #229 by @dgarijo plus the following changes:

1. Merge conflicts fixed
2. Updated the example to be compacted with the Codemeta context
3.  normalized the formatting in the example (indents, newlines)
4. Removed mention of being the reverse of `targetProduct`
 
Addresses #198 and #246


dgarijo:
@progval thanks a lot. I dropped the ball on the conflict resolving. You are a star!

progval:
I just added the reverse property `isSourceCodeOf`.

The name follows the schema.org conventions for reverse properties (eg. hasPart/isPartOf, hasVariant/isVariantOf)

dgarijo:
We may be asked about the difference between isSourceCodeOf and targetProduct. Both link a `ScoftwareSourceCode` with the `Software` it aims to represent no? I support `hasSourceCode` (hence this issue). But I can't think of a great difference between isSourceCodeOf and targetProduct.

The current definition is `Link that states where software application is built from a given source code`. Note that not all software is built from code, as interpreted code like Python does not need to be built. 

`targetProduct` is defined as `Target Operating System / Product to which the code applies. If applies to several versions, just the product name can be used`.  The definitions are quite similar, no?

Should `targetProduct` be the inverse of `hasSourceCode` instead?

progval:
I feel `targetProduct` is closer to `runtimePlatform`. In software engineering, one would say their app targets Windows, Flash, or the Web, to mean what product they aim to support, in the same way one targets an audience or democraphic. The schema.org definition implies it by giving Operating System as an example.

I've never seen someone says their source code targets the very app produced from their code. Though I admit it's common for binaries to be referred to as targets of the code (eg. in the context of a build system).

Either way, if we do come to the conclusion that `targetProduct` is the reverse of `hasSourceCode`, then `hasSourceCode` should be renamed `targetProductOf` for consistency.

dgarijo:
Hmmm, I saw `targetProduct` as the compiled binaries created by the source code (e.g., the JAR in Java), which  would then be considered `SoftwareApplication`s. Otherwise I am not sure what the difference with `runningPlatform` is.

I would keep `hasSourceCode` as it feels more natural than `targetProductOf`, to be honest (`targetProduct` is already a little bit confusing and per the documentation it may also be interpreted as the OS supporting the release).

If we want to keep `isSourceCodeOf`, then I feel like we need to justify why is it different to `targetProductOf`. Any takers to introduce a difference? Something like "this is not just for the binary, but to the abstract concept of software"?

moranegg:
A very interesting discussion.

`targetProduct` = Target Operating System / Product to which the code applies. If applies to several versions, just the product name can be used.
`runtimePlatforme` = Runtime platform or script interpreter dependencies (Example - Java v1, Python2.3, .Net Framework 3.0). Supersedes runtime.

If we take a simple example of the CodeMeta generator, you have the hosted instance: https://codemeta.github.io/codemeta-generator/

What would that be a `targetProduct` or `isSourceCodeOf`? 
Another possibility is to use `targetProduct`= Python's HTTP server
And the `runtimePatforme`= Python3 (or be more specific)

If this example doesn't work, we can try finding other examples were this can apply.
The idea here is to provide expressiveness following the issue #198 

At this stage we really want to release a V3.0 (after three years of intention).
Let's add both properties as suggested here and start a discussion for better examples in the discussion tab.
This might yield a larger discussion on relationships between objects, which is not clear how to express in CodeMeta.

progval:
Sounds fair. `targetProduct` is very unclear (as evidence by all the discussions around it), so if anything, `isSourceCodeOf` fixes the ambiguity.

PR #259, opened 2021-06-30
change embargoDate to embargoEndDate

tmorrell:
Per discussion in https://github.com/codemeta/codemeta/issues/254

PR #301, opened 2023-05-25
Deprecate `embargoDate` and replace it with `embargoEndDate`

moranegg:
Includes modifications for the new property embargoEndDate:
- Added to jsonld
- Added to properties
- Added to all crosswalk files


progval (approved):

progval:
we're going to go with https://github.com/codemeta/codemeta/pull/259 instead, given that we are going to skip the deprecation step (see https://github.com/codemeta/codemeta/discussions/306 )

PR #322, opened 2023-10-10
Remove restriction of 'identifier' to URIs

progval:
In practice, it is used to store non-URI identifiers, such as HAL IDs or (arguably ambiguous) names like here:

https://github.com/codemeta/codemeta/blob/7e51f02426f1a6d9e86edb3e99024dd0999ef97e/codemeta.json#L4

### DIFF
--- a/codemeta.jsonld
+++ b/codemeta.jsonld
@@ -36,7 +36,7 @@
       "funder": { "@id": "schema:funder"},
       "givenName": { "@id": "schema:givenName"},
       "hasPart": { "@id": "schema:hasPart" },
-      "identifier": { "@id": "schema:identifier", "@type": "@id"},
+      "identifier": { "@id": "schema:identifier" },
       "installUrl": { "@id": "schema:installUrl", "@type": "@id"},
       "isAccessibleForFree": { "@id": "schema:isAccessibleForFree"},
       "isPartOf":  { "@id": "schema:isPartOf"},


moranegg:
Is this `breaking` or `non-breaking` as it touches the vocabulary but seems like a change that will open the scope for identifier, I'm not sure.

dgarijo:
I think it is not breaking. It relaxes the constraints of the vocabulary

progval:
I'd say breaking because this document:

```json
{
  "http://schema.org/identifier": "foo bar"
}
```

currently gets compacted to:

```json
{
  "@context": "https://w3id.org/codemeta/3.0",
  "schema:identifier": "foo bar"
}
```

but after the change it would get compacted to:

```json
{
  "@context": "https://w3id.org/codemeta/3.0",
  "identifier": "foo bar"
}
```

> I think it is not breaking. It relaxes the constraints of the vocabulary

Relaxing contraints of a format is a breaking change for readers. (And adding constraints is a breaking change for writers.)

dgarijo:
sorry, why would it remove/add a namespace? The only thing the id is doing is expecting an object, no?
It would be a problem for a validator, I agree. But the property name does not change with the change.

progval:
because `"foo bar"` is not a URI, so it's not a valid `@id`. Which means the `"identifier": { "@id": "schema:identifier", "@type": "@id"},` rule in the context can't be used to replace `schema:identifier` with `identifier`.

dgarijo:
But identifier is mapped to schema:identifier per the context file...

dgarijo:
right, so since we are relaxing the constraint, everything that worked before the change, will continue to work. Unless you have specified a test for your validator against that particular constraint... In any case, if you think this is something that should be delayed then ok.

progval:
What validator are you talking about? This changes the behavior of any spec-compliant implementation of the JSON-LD compaction algorithm.

progval:
Actually, it changes the behavior of the expansion algorithm too:

```json
{
  "@context": {
    "schema": "http://schema.org",
    "identifier": {"@id": "schema:identifier", "@type": "@id"}
  },
  "identifier": "foo bar"
}
```

is expanded to:

```json
[
  {
    "schema:identifier": [
      {
        "@id": "foo bar"
      }
    ]
  }
]
```

while 

```json
{
  "@context": {
    "schema": "http://schema.org",
    "identifier": {"@id": "schema:identifier"}
  },
  "identifier": "foo bar"
}
```

is expanded to

```json
[
  {
    "schema:identifier": [
      {
        "@value": "foo bar"
      }
    ]
  }
]
```

as well as the semantics of the document (eg. when loaded in RDFlib), because the value is now a literal (`@value`) instead of a URI (`@id`).

dgarijo:
I did not realize about the resource changing its type from Resource (or Thing) to Literal, that is from:
```
_:b0 schema:identifier <https://example.org> .
```
to 
```
_:b0 schema:identifier "https://example.org" .
```
Apologies! I got confused because I thought it was a namespace issue instead of a change of type in the literal.

tmorrell:
This is included in the draft v4 at https://github.com/SciCodes/codemeta/blob/target_codemeta_v4/codemeta.jsonld

bzg:
@moranegg can you close this PR? The change is redundant with some of the changes in https://github.com/codemeta/codemeta/pull/486, where I believe the review should happen.

PR #324, opened 2023-10-19
[wikidata] Replace issue tracker URL

physikerwelt:
issue tracker URL (P1401) https://www.wikidata.org/wiki/Property:P1401 is the wikidata property associated with issue tracking


progval:
Thanks!

progval (approved):

PR #279, opened 2022-08-10
Create MODS.csv

avandenbosch:
PR for a first attempt at a MODS-codemeta crosswalk.

PR #326, opened 2023-10-19
fix: adjust maven mappings for codemeta v3

alee:
bring Java/Maven crosswalk mapping to codemeta v3 and adjust some of the original mappings to the Maven Project Object Model

https://maven.apache.org/pom.html

- codeRepository should point at the place where the human readable uncompiled source code lives so `scm`
(https://maven.apache.org/pom.html#scm) is the right choice, not repositories (https://maven.apache.org/pom.html#repositories)
- softwareRequirements is for the source code's dependencies (https://maven.apache.org/pom.html#dependencies), not the maven pom.xml's prerequisites (https://maven.apache.org/pom.html#prerequisites) 
- one issue on dependencies is that Maven does transitive dependency management so the full set of dependencies is not resolved in this pom but would need to be programmatically generated (e.g., `mvn dependency:tree` or `mvn dependency:list`)

### DIFF
--- a/crosswalks/Java (Maven).csv
+++ b/crosswalks/Java (Maven).csv
@@ -1,22 +1,22 @@
 Property,Java (Maven)
-codeRepository,scm
+codeRepository,scm / url
 programmingLanguage,
 review,
 runtimePlatform,
 targetProduct,
 applicationCategory,
 applicationSubCategory,
-downloadUrl,distributionManagement
+downloadUrl,distributionManagement / downloadUrl
 fileSize,
 installUrl,
 memoryRequirements,
 operatingSystem,
 permissions,
 processorRequirements,
-releaseNotes,description
+releaseNotes,
 softwareHelp,
 softwareRequirements,dependencies
-softwareVersion,
+softwareVersion,version
 storageRequirements,
 supportingData,
 author,developers
@@ -32,7 +32,7 @@ encoding,
 fileFormat,
 funder,
 keywords,
-license,license
+license,licenses / license / url
 producer,
 provider,
 publisher,
--- a/crosswalks/Java (Maven).csv
+++ b/crosswalks/Java (Maven).csv
@@ -1,30 +1,30 @@
 Property,Java (Maven)
-codeRepository,scm / repositories
+codeRepository,scm
 programmingLanguage,
 review,
 runtimePlatform,
 targetProduct,
 applicationCategory,
 applicationSubCategory,
-downloadUrl,
+downloadUrl,distributionManagement
 fileSize,
 installUrl,
 memoryRequirements,
 operatingSystem,
 permissions,
 processorRequirements,
-releaseNotes,
+releaseNotes,description
 softwareHelp,
-softwareRequirements,prerequisites
+softwareRequirements,dependencies
 softwareVersion,
 storageRequirements,
 supportingData,
-author,
+author,developers
 citation,
-contributor,
+contributor,contributors
 copyrightHolder,
 copyrightYear,
-dateCreated,
+dateCreated,inceptionYear
 dateModified,
 datePublished,
 editor,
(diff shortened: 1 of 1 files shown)


progval:
Hi,

Thanks for the fixes. Could you undo changes in line ends to keep the diff minimal?

> Maven does transitive dependency management so the full set of dependencies is not resolved

This isn't an issue, most package managers work like this

crosswalks/Java (Maven).csv
```diff
@@ -1,30 +1,30 @@
...
-downloadUrl,
+downloadUrl,distributionManagement
 fileSize,
 installUrl,
 memoryRequirements,
 operatingSystem,
 permissions,
 processorRequirements,
-releaseNotes,
+releaseNotes,description
```
progval, line 16: Are you sure? https://maven.apache.org/pom.html#More_Project_Information says `description` is "A short, human readable description of the project."

crosswalks/Java (Maven).csv
```diff
@@ -1,30 +1,30 @@
...
-codeRepository,scm / repositories
+codeRepository,scm
 programmingLanguage,
 review,
 runtimePlatform,
 targetProduct,
 applicationCategory,
 applicationSubCategory,
-downloadUrl,
+downloadUrl,distributionManagement
```
progval, line 9: Should be `distributionManagement/downloadUrl` according to https://maven.apache.org/pom.html#distribution-management

crosswalks/Java (Maven).csv
```diff
@@ -1,30 +1,30 @@
 Property,Java (Maven)
-codeRepository,scm / repositories
+codeRepository,scm
```
progval, line 2: Should be `scm/url` according to https://maven.apache.org/pom.html#SCM

progval (changes requested):

crosswalks/Java (Maven).csv
alee: :+1: probably reasonable to leave it out since I think we already map description -> description earlier

crosswalks/Java (Maven).csv
alee: Are there docs on how these crosswalks are intended to handle mappings to nested metadata? Since these crosswalks aren't even machine actionable I figured it'd be easier to point at the parent element.

I don't have the bandwidth to devote to refining these further, so if anyone else wants to contribute those changes please feel free!

progval: Not really, but in #279 I (dictatorially) decided to use an XPath-like syntax given the somewhat complex matching requirements, and I figure we might as well be consistent with that decision here.

> Since these crosswalks aren't even machine actionable

At Software Heritage we kind of use machine-read them: https://gitlab.softwareheritage.org/swh/devel/swh-indexer/-/blob/master/swh/indexer/metadata_dictionary/maven.py

Though of course there's a lot of Python code needed to validate and re-shape the data, but the XPath format may help

crosswalks/Java (Maven).csv
alee: How about `scm / url` which should be the landing page to the code repository instead of the clone / checkout url

progval: Hmm, https://schema.org/codeRepository is ambiguous about which it should be. However "http://127.0.0.1/websvn/my-project" looks way more usable than "scm:svn:http://127.0.0.1/svn/my-project" (which seems to be a maven-ism) so let's go for `scm / url`

crosswalks/Java (Maven).csv
alee: `license` should probably go to `licenses / license` or `licenses / license / url` then?

https://maven.apache.org/pom.html#licenses

progval: `licenses / license / url`, yes

PR #327, opened 2023-10-19
DOAP revision

dgarijo:
This push has a revision of the DOAP crosswalk:

- Organized the codemeta properties aphabetically to ease mapping
- Added a new column with comments and rationale
- Identifier properties from doap that map to the same codemeta property. Examples: `doap:shortDesc` and `doap:description` map to `schema:description`, several contributor roles in DOAP map to schema:contributor


dgarijo:
I guess adding a mapping rationale in the CSV breaks the validator. Is there a way to fix this @progval ? I can change the PR to remove the third column and maybe add this in a different file?
The comments are important to understand the context of the mapping.

dgarijo:
@moranegg adding a third column in the csv breaks the validator. So I added the rationale in a README file, following the example of CFF. I know you don't like this, but I don't have an alternative for now :(

moranegg:
Thanks for trying but...
It is impossible to maintain having two documents trying to say the same thing.
When I thought of a README, it was to add some metmetadata about the provenance of the crosswalk and not another mapping in .md format.

Let's try and find a more sustainable solutions even if we don't have the "good" info now.

please delete the README, and add this information not as a table but as remarks in an issue.

thanks again for your effort :-)

dgarijo:
Then, shall we also delete the CFF readme?

progval:
> Then, shall we also delete the CFF readme?

The CFF readme only gives rationale behind choices, not a copy of the table. You can do the same here

dgarijo:
doap:repository is property linking doap:Projects with doap:Repositories. Both are object, not URLs. 

doap:repository does not link to the URL, if we consider a doap:Project a SoftwareApplication, my proposal stands.

progval:
Oh wait, I get it. `codeRepository and address should be the same` is actually caused by you changing the order of entries, and the CI expects the first property to be `codeRepository`

dgarijo:
@progval I have checked more of their examples, and here is evidence of my point:
```:repository [
        :browse <https://github.com/ewilderj/doap/> ;
        :location <https://github.com/ewilderj/doap.git> ;
        a :GitRepository
    ] ;
```
The range of `repository` is supposed to be "SoftwareSourceCode", which is reange of `hasSourceCode`. They also have the inverse property declares `repositoryOf`. I added `location` as mapping to schema: codeRepository

moranegg:
Yes please revert to the original order, this PR won’t match the script aggregating the mappings into the crosswalk table.

dgarijo:
Ok, I will revert to the original unordered file when I get the chance

moranegg:
Hi @dgarijo  we are moving forward with v4.0 .
If you want this PR to be in v4.0, please amend PR before January 10th 2024.
We will be very glad to have it in the next release if the files follows the codemeta crosswalk format.

dgarijo:
I will amend the order this week.

dgarijo:
I have reverted and removed the "comments" section. Since there is no other place where to put the rationale, I will do so here. The rationale has source property, target property and comments explaining the rationale:
- applicationCategory,category,exact match
- applicationSubCategory,category,broader than
- author,developer,broader than
- codeRepository,location, removed `repository` as mapping, because the range of repository is SourceCode, not the URL to the git repo. Location is used for that
- contributor,developer,"doap:documenter, doap:translator, doap:tester, doap:helper also map to this category"
- dateCreated,created,exact match
- description,description,doap:shortdesc also maps to description (narrower than)
- downloadUrl,download-page,possible map to doap:download-mirror as well. doap:file-release is narrower than downloadUrl (URL where to download the release)
- givenName,,Not mapped because this property is for persons
- hasSourceCode,repository,doap:repository is property linking doap:Projects with doap:Repositories
- installUrl,download-mirror,
- isSourceCodeOf,repositoryOf,inverse of repository
- issueTracker,bug-database,
- keywords,category,
- license,license,exact match
- maintainer,maintainer,exact match
- name,name,exact match
- operatingSystem,os,exact match
- programmingLanguage,programming-language,exact match
- publisher,vendor,
- referencePublication,blog,related
- runtimePlatform,platform,exact match
- softwareHelp,support-forum,developer-forum also maps. And doap:wiki
- softwareVersion,release,related. A release in doap links a project to a doap:Version. In schema.org this is more like the revision
- url,homepage,exact match. doap:old-homepage also maps to this propety (url is broader than). doap:location also maps to this property (for codeRepositories)
- version,revision,exact match

progval:
I'm not sure it is correct to map `doap:repository` to `codemeta:hasSourceCode` because the former links to a repository and the latter links to source code, which are not the same object (a repository contains source code); as you pointed out they are even formally defined as being different types.

Also, given that `doap:location` is nested inside `doap:repository`, you might want to replace the first line with `codeRepository,repository/location`

dgarijo:
I will add the repo/location.

dgarijo:
It's a mapping, so none of the objects will be exactly equivalent. But the
object itself (source code) has some sort of equivalence, in this case
pointing to the object denoting the code itself. I think it is ok, since
it's explained in the rationale. If you find a counter example where the
relationship would be different we can always go back.

dgarijo:
@progval I added your suggestions.
Since I don't really mind fighting the repository/repositoryOf mapping, I have removed it, in case it helps merging the PR before the next release.

progval (approved):

dgarijo:
@progval thanks!

PR #341, opened 2024-01-06
Update properties_description.csv - add quotes around the last field

bact:
Line 60, last field:

The field contains comma `,` which can confuse CSV parser and thought that the row has 5 columns instead of 4.


bact:
See #342 instead

PR #342, opened 2024-01-06
Fix CSV column inconsistency and roleName type

bact:
## Fix CSV column inconsistency and roleName type
- Line 60 and 62: Add quotes around the last field (which contains a comma `,` that can confused CSV parser and thought that the rows have 5 columns instead of 4)
- Line 61: `roleName` should be `Text or URL` (currently it has Text followed by URL - `Text,URL`, which also make it 5 columns)

By fixing these column number inconsistencies, GitHub will able to display the file as a table.

## Format/description consistency
- Line 65, 66, 70, 72: Capitalizes first character in the sentence.
- Put "a" determiner for "link", make it "a link" just as "a license", "a flag", etc. as appeared in other property descriptions.
- Replace double spaces with single spaces

### DIFF
--- a/properties_description.csv
+++ b/properties_description.csv
@@ -58,7 +58,7 @@ schema:Person,address,PostalAddress or Text,Physical address of the item.
 schema:Review,reviewAspect,Text,This review is relevant to this part or facet of the object being review
 schema:Review,reviewBody,Text,The actual body of the review
 schema:Role,endDate,Date or Datetime,"The end date and time of the item, usually a Role. In ISO 8601 date format"
-schema:Role,roleName,Text,"A role played, performed, or filled by a person or organization. Suggested values: ""Design"" (algorithm), ""Architecture"", ""Debugging"", ""Maintenance"", ""Coding"", ""Documentation"", ""Testing"", ""Support"", ""Management""."
+schema:Role,roleName,Text or URL,"A role played, performed, or filled by a person or organization. Suggested values: ""Design"" (algorithm), ""Architecture"", ""Debugging"", ""Maintenance"", ""Coding"", ""Documentation"", ""Testing"", ""Support"", ""Management""."
 schema:Role,startDate,Date or Datetime,"The start date and time of the item, usually a Role. In ISO 8601 date format"
 codemeta:SoftwareSourceCode,softwareSuggestions,SoftwareSourceCode,"Optional dependencies, e.g. for optional features, code development, etc."
 codemeta:SoftwareSourceCode,maintainer,Person,Individual responsible for maintaining the software (usually includes an email contact address)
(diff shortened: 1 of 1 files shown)


properties_description.csv
```diff
@@ -55,28 +55,21 @@ schema:Person,affiliation,Organization,"An organization that this person is affi
...
+schema:Review,reviewAspect,Text,This review is relevant to this part or facet of the object being review
 schema:Review,reviewBody,Text,The actual body of the review
-schema:Role,endDate,Date or Datetime,The end date and time of the item, usually a Role. In ISO 8601 date format
-schema:Role,roleName,Text,URL,"A role played, performed, or filled by a person or organization. Suggested values: ""Design"" (algorithm), ""Architecture"", ""Debugging"", ""Maintenance"", ""Coding"", ""Documentation"", ""Testing"", ""Support"", ""Management""."
-schema:Role,startDate,Date or Datetime,The start date and time of the item, usually a Role. In ISO 8601 date format
-codemeta:SoftwareSourceCode,softwareSuggestions,SoftwareSourceCode,"Optional dependencies , e.g. for optional features, code development, etc."
+schema:Role,endDate,Date or Datetime,"The end date and time of the item, usually a Role. In ISO 8601 date format"
+schema:Role,roleName,Text,"A role played, performed, or filled by a person or organization. Suggested values: ""Design"" (algorithm), ""Architecture"", ""Debugging"", ""Maintenance"", ""Coding"", ""Documentation"", ""Testing"", ""Support"", ""Management""."
```
progval, line 61:
```suggestion
schema:Role,roleName,Text or URL,"A role played, performed, or filled by a person or organization. Suggested values: ""Design"" (algorithm), ""Architecture"", ""Debugging"", ""Maintenance"", ""Coding"", ""Documentation"", ""Testing"", ""Support"", ""Management""."
```

properties_description.csv
```diff
@@ -55,28 +55,21 @@ schema:Person,affiliation,Organization,"An organization that this person is affi
...
-,,,
-,,,
-,,,
-,,,
-,,,
-,,,
+codemeta:SoftwareSourceCode,readme,URL,A link to software Readme file
+codemeta:SoftwareApplication,hasSourceCode,SoftwareSourceCode,A link that states where the software code is for a given software. For example a software registry may indicate that one of its software entries hasSourceCode in a GitHub repository.
+codemeta:SoftwareSourceCode,isSourceCodeOf,SoftwareApplication,A link that states where software application is built from a given source code. This is the reverse property of 'hasSourceCode'.
 ,,,
```
progval, line 75:
```suggestion
```

progval (changes requested):
Much needed fixes, thank you. Two suggestions below

properties_description.csv
bact: Thank you. Should I make this one `Text or URL` instead then?

bact:
Thank you for suggestions. I have made commits for all suggestions.

progval (approved):

bact:
btw, roleName in https://github.com/codemeta/codemeta/blob/master/crosswalk.csv is `Text` without `URL`.

So we now have an inconsistency between two files:
- https://github.com/codemeta/codemeta/blob/master/properties_description.csv
- https://github.com/codemeta/codemeta/blob/master/crosswalk.csv

progval:
We periodically regenerate `crosswalk.csv` from `properties_description.csv` and others.

bact:
Got it. So it will be eventually consistent, thank you.

PR #344, opened 2024-01-09
Add SPDX 2.3 crosswalk

bact:
Add SPDX 2.3 crosswalk

ref: https://spdx.github.io/spdx-spec/v2.3/


progval:
Could you submit changes unrelated to SPDX as separate pull-requests, and remove them for this one?

> Remove excessed white spaces from start and end of fields

Specifically, they are carriage-returns, we keep mixing Windows-style (`\r\n`) and Unix-style (`\n` line ends) in these files :/

bact:
> Could you submit changes unrelated to SPDX as separate pull-requests, and remove them for this one?

Will do that. I did that in the first place because the aggregate script doesn't generate last three rows for SPDX (and others).

I will make a separated PR.

bact:
@progval for `SciCrunchRegistry.csv` do you like to keep the Unicode Character 'NO-BREAK SPACE' (U+00A0)?

They started to appear after line 20 in the file.
(Does not appear in the first 20 lines)

crosswalks/SPDX_2_3.csv
```diff
@@ -0,0 +1,74 @@
...
+endDate,
+roleName,
+startDate,
+softwareSuggestions,
+maintainer,
+continuousIntegration,
+buildInstructions,
+developmentStatus,
+embargoEndDate,
+funding,PackageAttributionText
```
progval, line 69: This doesn't seem like a very good match as `PackageAttributionText` is for free-form acknowledgement, but I don't see a better option. Should we keep it?

crosswalks/SPDX_2_3.csv
```diff
@@ -0,0 +1,74 @@
+Property,SPDX 2.3
+codeRepository,PackageSourceInfo
+programmingLanguage,
+review,Annotator
```
progval, line 4:
```suggestion
review,Annotation
```

`Annotator` is for the author of the review, not the review itself.

crosswalks/SPDX_2_3.csv
bact: May be it's safer to remove it.

crosswalks/SPDX_2_3.csv
```diff
@@ -0,0 +1,74 @@
...
+relatedLink,
+givenName,
+familyName,
+email,
+affiliation,
+identifier,
+name,PackageOriginator
+address,
+reviewAspect,AnnotationType
+reviewBody,Annotator
```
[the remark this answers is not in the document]
bact: Thank you. You're right. I was misread it. I have updated it accordingly.

crosswalks/SPDX_2_3.csv
bact: I have removed it.

progval (approved):
Looks good to me, thanks! I'll merge this next week if no one else wants to comment.

bact:
Thank you.
I will share this PR with SPDX community, in case they have more comments.

PR #363, opened 2024-07-17
Add BibTeX mapping

### LINKED ISSUE #58: add mapping to bibtex software entries

hjonin:
Resolves #58.

### DIFF
--- a/crosswalks/BibTeX.csv
+++ b/crosswalks/BibTeX.csv
@@ -16,7 +16,7 @@ processorRequirements,
 releaseNotes,
 softwareHelp,
 softwareRequirements,
-softwareVersion,version (1)
+softwareVersion,version
 storageRequirements,
 supportingData,
 author,author
@@ -24,9 +24,9 @@ citation,
 contributor,
 copyrightHolder,
 copyrightYear,
-dateCreated,date (2) (year) / (month)
-dateModified,date (3) (year) / (month)
-datePublished,date (1) (year) / (month)
+dateCreated,
+dateModified,
+datePublished,date / (year) / (month)
 editor,editor
 encoding,
 fileFormat,
@@ -37,7 +37,7 @@ producer,
 provider,
 publisher,publisher
 sponsor,
-version,version (2)
+version,
 isAccessibleForFree,
 isPartOf,
 hasPart,
@@ -48,12 +48,12 @@ name,title
 sameAs,
 url,url
 relatedLink,
-givenName,author
-familyName,author
+givenName,
+familyName,
 email,
 affiliation,organization
 identifier,
-name,author
+name,
 address,
 reviewAspect,
 reviewBody,
--- a/crosswalks/BibTeX.csv
+++ b/crosswalks/BibTeX.csv
@@ -13,7 +13,7 @@ memoryRequirements,
 operatingSystem,
 permissions,
 processorRequirements,
-releaseNotes,
+releaseNotes,note
 softwareHelp,
 softwareRequirements,
 softwareVersion,version
--- a/crosswalks/BibTeX.csv
+++ b/crosswalks/BibTeX.csv
@@ -1,0 +1,74 @@
+Property,BibTeX (@softwareversion)
+codeRepository,repository
+programmingLanguage,
+review,
+runtimePlatform,
+targetProduct,
+applicationCategory,
+applicationSubCategory,
+downloadUrl,file
+fileSize,
+installUrl,
+memoryRequirements,
+operatingSystem,
+permissions,
+processorRequirements,
+releaseNotes,
+softwareHelp,
+softwareRequirements,
+softwareVersion,version (1)
+storageRequirements,
+supportingData,
+author,author
+citation,
+contributor,
+copyrightHolder,
+copyrightYear,
+dateCreated,date (2) (year) / (month)
+dateModified,date (3) (year) / (month)
+datePublished,date (1) (year) / (month)
+editor,editor
+encoding,
+fileFormat,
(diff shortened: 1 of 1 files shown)


crosswalks/bibtex.csv
```diff
@@ -0,0 +1,74 @@
...
+fileSize,
+installUrl,
+memoryRequirements,
+operatingSystem,
+permissions,
+processorRequirements,
+releaseNotes,
+softwareHelp,
+softwareRequirements,
+softwareVersion,version (1)
```
hjonin, line 19: I added a number here to indicate priority of fields, but it might not be consistent with other mappings, so I can remove it.

crosswalks/BibTeX.csv
```diff
@@ -0,0 +1,74 @@
+Property,BibTeX (@softwareversion)
```
hjonin, line 1: I have indicated the entry type like this, does it seem ok?

crosswalks/BibTeX.csv
```diff
@@ -0,0 +1,74 @@
...
+softwareRequirements,
+softwareVersion,version (1)
+storageRequirements,
+supportingData,
+author,author
+citation,
+contributor,
+copyrightHolder,
+copyrightYear,
+dateCreated,date (2) (year) / (month)
```
hjonin, line 27: I have indicated the priority of fields with a number, but I can remove it if it is not consistent with other mappings.

crosswalks/BibTeX.csv
```diff
@@ -0,0 +1,74 @@
...
+name,title
+sameAs,
+url,url
+relatedLink,
+givenName,author
+familyName,author
+email,
+affiliation,organization
+identifier,
+name,author
```
progval, line 56: given we also map `author,author`, this implies that "givenName of author" in Codemeta matches "author of author" in BibTeX, which does not really make sense. I believe we should remove these three name mappings, and leave crosswalks users to figure it out.

crosswalks/BibTeX.csv
```diff
@@ -0,0 +1,74 @@
...
+storageRequirements,
+supportingData,
+author,author
+citation,
+contributor,
+copyrightHolder,
+copyrightYear,
+dateCreated,date (2) (year) / (month)
+dateModified,date (3) (year) / (month)
+datePublished,date (1) (year) / (month)
```
progval, line 29: https://www.bibtex.com/f/year-field/ says:

> The year field is used to store the year the work was published or in the case of an unpublished article the year it was written.

schema.org/codemeta has no concept of "the date it was written", so in my opinion, this should only map to `datePublished`, not `dateCreated` or `dateModified`.

crosswalks/BibTeX.csv
progval: See above for the multiplicity of dates. For versions, honestly I don't know why schema.org defines both `version` and `softwareVersion`.

Other crosswalks just pick one (usually `softwareVersion`) or sometimes both, so you should do like they do.

We should have a policy for this eventually, but that's out of scope of this PR.

hjonin: For `softwareVersion`, I did like [CFF mapping](https://github.com/codemeta/codemeta/blob/3c30d4cb59864e9d89ee607edd1e5382ee582704/crosswalks/Citation%20File%20Format%201.2.0.csv#L19) which maps both, but I agree to just keep `softwareVersion`.

crosswalks/BibTeX.csv
hjonin: I got his mapping from [the codemeta to bibtex algorithm](https://gitlab.softwareheritage.org/swh/devel/swh-indexer/-/blob/master/swh/indexer/bibtex.py?ref_type=heads#L86). Do you think I should update the algorithm as well?

crosswalks/BibTeX.csv
hjonin: I am not sure to understand what you mean here, would you have any example?
Do you mean that we should not map `author` at all?

progval: I mean that `author` should be mapped, but not `givenName`, `familyName`, or `name`

crosswalks/BibTeX.csv
progval: It's fine for swh-indexer to differ from the crosswalks, because crosswalks can't represent anywhere near as much complexity as actual code.

hjonin: Ok!

crosswalks/BibTeX.csv
moranegg: +1

crosswalks/BibTeX.csv
```diff
@@ -0,0 +1,74 @@
...
+applicationCategory,
+applicationSubCategory,
+downloadUrl,file
+fileSize,
+installUrl,
+memoryRequirements,
+operatingSystem,
+permissions,
+processorRequirements,
+releaseNotes,
```
moranegg, line 16: add note (see https://gitlab.inria.fr/gt-sw-citation/bibtex-sw-entry/-/blob/master/swentry.org?ref_type=heads)

moranegg (commented):
Thank @hjonin for the crosswalk!

Can you add th `note` mapping to `releaseNotes`?

With this change we can approve and integrate the crosswalk.

hjonin:
> Thank @hjonin for the crosswalk!
> 
> Can you add th `note` mapping to `releaseNotes`?
> 
> With this change we can approve and integrate the crosswalk.

Done in the last commit :)

Issue #197, opened 2018-10-11
Potential vocabularies to crosswalk


cboettig:
To avoid a collection of stale issues on which additional vocabularies we might add to the crosswalk, I'll try and aggregate a running list into the thread here.  At this time we don't have developer capacity to add crosswalks ourselves but PRs always welcome.  

- [ ] .ABOUT (was #152)
- [ ] ADMS.SW (#41)
- [ ] seon (#127)
- [ ] debtags (#127)
- [ ] jats (#48)
- [ ] bibtex (#58)
- [ ] DCAT (#44)
- [ ] DIRT (#119)
- [ ] ....
Arguably not all of these should be crosswalked, in some cases something more machine-readable such as an ontology specification could provide a more practical mechanism.

moranegg:
Great idea to list all vocabularies that can/should be crosswalked in the same issue.
I would add software registries that have metadata about software, including:
- [ ] libraries.io
- [ ] swMath 
- [ ] ascl.net 

I don't know if each uses an existing vocabulary, but i think it is important to verify and crosswalk with CodeMeta.

moranegg:
I want to do a PR for BibTeX with the following BibLaTeX software style.
https://www.ctan.org/tex-archive/macros/latex/contrib/biblatex-contrib/biblatex-software
The terms are detailed here:
https://gitlab.inria.fr/gt-sw-citation/bibtex-sw-entry
With 4 different software levels/types:
- @software, 
- @softwareversion,
- @softwaremodule, 
- @codefragment

I tend to think that each type should have a column.
Should I do one .csv file with five columns (CodeMeta, software, softwareversion, softwaremodule, codefragment)?

mbjones:
@moranegg I suspect they should be separate as well given that each of the bibtex types has a different set of allowed fields.  The 4 crosswalks will likely be very similar, but each will have its own unique fields. So +1 for separate mappings, which is how they will be encountered in the wild as well.

hjonin:
I would add the vocabulary:

- [ ] [publiccode.yml](https://github.com/publiccodeyml/publiccode.yml)

PR #364, opened 2024-07-17
Add publiccode.yml mapping

hjonin:
See https://docs.italia.it/italia/developers-italia/publiccodeyml-en/en/master/schema.core.html#top-level-keys-and-sections and https://github.com/codemeta/codemeta/issues/197#issuecomment-2205461293.


crosswalks/publiccode.csv
```diff
@@ -0,0 +1,74 @@
...
+continuousIntegration,
+buildInstructions,
+developmentStatus,developmentStatus
+embargoEndDate,
+funding,
+issueTracker,
+referencePublication,
+readme,
+hasSourceCode,
+isSourceCodeOf,isBasedOn
```
progval, line 74: isSourceCodeOf is to describe relationships between an application and its code. https://schema.org/isBasedOn is what you are looking for, but it's not part of Codemeta.

crosswalks/publiccode.csv
```diff
@@ -0,0 +1,74 @@
+Property,publiccode
+codeRepository,url
+programmingLanguage,
+review,
+runtimePlatform,platforms
```
progval, line 5: https://docs.italia.it/italia/developers-italia/publiccodeyml-en/en/master/schema.core.html#key-platforms says:

> This key specifies which platform the software runs on. It is meant to describe the platforms that users will use to access and operate the software, rather than the platform the software itself runs on.

this seems to explicitly exclude mapping to runtimePlatform.

crosswalks/publiccode.csv
```diff
@@ -0,0 +1,74 @@
...
+targetProduct,platforms
```
progval, line 6: I was about to debate this, because (quoting [this old comment of mine](https://github.com/codemeta/codemeta/pull/300#issuecomment-1619960326)):

> In software engineering, one would say their app targets Windows, Flash, or the Web, to mean what product they aim to support, in the same way one targets an audience or democraphic.

However, given the description in https://docs.italia.it/italia/developers-italia/publiccodeyml-en/en/master/schema.core.html#key-platforms I think it might actually be relevant.

crosswalks/publiccode.csv
```diff
@@ -0,0 +1,74 @@
...
+applicationCategory,categories / description.[lang].genericName
+applicationSubCategory,categories
+downloadUrl,landingUrl
+fileSize,
+installUrl,landingUrl
```
progval, line 11: The landing URL is not machine-readable, so I think we should not map these properties to it.

crosswalks/publiccode.csv
hjonin: Ok! This description is in fact very confusing to me because it seems to say 2 contradictory things (`the platform the software runs on` and `not the platform the software runs on`. :smile:

crosswalks/publiccode.csv
hjonin: Do you mean the target is not readable or `landingUrl` itself is not machine readable? Because, if the latter, it should, since it's a URL :slightly_smiling_face:

crosswalks/publiccode.csv
progval: Hah, indeed. I missed the first sentence was the exact opposite of the second one...

crosswalks/publiccode.csv
progval: I mean the target

bzg:
Anything we can do to progress on this PR?

progval (approved):

bzg:
Thanks:

bzg:
Is there anything I can do to help review/complete/merge this PR?

Thanks!

progval:
Sorry, I was giving other people to give feedback and then forgot. Merged now

bzg:
Thanks a lot!

PR #366, opened 2024-09-27
VIVO crosswalk

chenejac:
# What does this pull request do?
Adds crosswalk between codemeta and the VIVO ontology software class and its object and data properties.

# What's new?
The VIVO.csv file which is describing the crosswalk. The second column besides the name of object/data property, specify  object/data  property URIs and range. Moreover, if some mapping is more complex and include more elements from the graph dot is used as separator in the following syntax:

funder, informationResourceSupportedBy.assignedBy (property URI - http://vivoweb.org/ontology/core#informationResourceSupportedBy; http://vivoweb.org/ontology/core#assignedBy; range URI - http://vivoweb.org/ontology/core#Grant; http://xmlns.com/foaf/0.1/Organization)

This mean information about funder organization might be found starting from a Software class through its object properties http://vivoweb.org/ontology/core#informationResourceSupportedBy which is linking a Software with a Grant, and the next level is to use http://vivoweb.org/ontology/core#assignedBy object property of the linked Grant to get Funder (an instance of http://xmlns.com/foaf/0.1/Organization).

In order to make this crosswalk, an analysis of the VIVO ontology and its alignment (extension) with codemeta was done. More details about extensions of the VIVO ontology can be found [here](https://github.com/vivo-ontologies/vivo-ontology/pulls?q=is%3Apr+label%3A%22software+codemeta+alignment%22)  

# Additional Notes:
The crosswalk will be validated in the practice via implementation of [a REST API endpoints in VIVO]( https://github.com/ivanmrsulja/VIVO/commits/feature/rsmd/) which will work with codemeta json data (which are mapped and preserved in the VIVO graph)

The work is supported by a small FAIR-Impact grant.

### DIFF
--- a/crosswalks/VIVO.csv
+++ b/crosswalks/VIVO.csv
@@ -1,4 +1,4 @@
-Property,VIVO (ontology URI - https://github.com/vivo-ontologies/vivo-ontology/blob/master/vivo.owl, Software main class URI - http://purl.obolibrary.org/obo/ERO_0000071)
+Property,VIVO (ontology URI - https://github.com/vivo-ontologies/vivo-ontology/blob/master/vivo.owl; Software main class URI - http://purl.obolibrary.org/obo/ERO_0000071)
 codeRepository,
 programmingLanguage,
 review,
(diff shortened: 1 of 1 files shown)


moranegg:
Thanks @chenejac!

@dgarijo, @progval: can you check this PR, I do see advantages of having the PURL in the mapping, but don't know if that is something the current format can support.

progval:
Thank you for providing value range, but we don't have them in other crosswalks so we should probably omit them here too, unless there is a specific reason I miss.

Regarding the property URIs, instead of providing both the local part (and what I assume is some sort of display-name for properties in the `http://purl.obolibrary.org/obo/`  namespace) and the URI, could you provide only the URI? (Or maybe a QName, like we do [for Dublin Core](https://github.com/codemeta/codemeta/blob/master/crosswalks/Dublin%20Core.csv)?)

chenejac:
@progval thanks for your comment.

We can remove local name, and use only URI, and even remove range URI. In that case, this line:

> funder,informationResourceSupportedBy.assignedBy (property URI - http://vivoweb.org/ontology/core#informationResourceSupportedBy; http://vivoweb.org/ontology/core#assignedBy; range URI - http://vivoweb.org/ontology/core#Grant; http://xmlns.com/foaf/0.1/Organization)

will be replaced with 

> funder, http://vivoweb.org/ontology/core#informationResourceSupportedBy.http://vivoweb.org/ontology/core#assignedBy

However, it is a question whether it is clear enough for people who would like to implement metadata transformation? 
It would be better if a crosswalk is expressed in machine-actionable format by using some formal mapping language and tool (for instance [MSCR](https://faircore4eosc.eu/eosc-core-components/metadata-schema-and-crosswalk-registry-mscr), [X3ML toolkit](https://www.ics.forth.gr/x3ml-toolkit), [RML](https://rml.io/specs/rml/), etc.). The CSV approach is not a formal notation for expressing mappings. In this concrete case, the information about funder for a software in VIVO can be found in the following way:

- there is an object property http://vivoweb.org/ontology/core#informationResourceSupportedBy linked with a Software instance
- range for this object property is a super class of http://vivoweb.org/ontology/core#Grant to ensure you can express that a software might be supported by some community, piece of equipment, grant, etc. Only in the case the object type is http://vivoweb.org/ontology/core#Grant, there might be http://vivoweb.org/ontology/core#assignedBy object property containing information about an Organization Unit representing a Funder. 

Therefore, in this scenario I believe information about range URI might be useful. Otherwise, I believe a potential implementer of the crosswalk transformation will have to investigate the VIVO ontology in more details. 

However, please select which one of the following four approaches we should follow in our PR, and we will update our PR in accordance with that decision:

1.   funder,informationResourceSupportedBy.assignedBy (property URI - http://vivoweb.org/ontology/core#informationResourceSupportedBy; http://vivoweb.org/ontology/core#assignedBy; range URI - http://vivoweb.org/ontology/core#Grant; http://xmlns.com/foaf/0.1/Organization)
2.  funder,http://vivoweb.org/ontology/core#informationResourceSupportedBy.http://vivoweb.org/ontology/core#assignedBy (range URI - http://vivoweb.org/ontology/core#Grant; http://xmlns.com/foaf/0.1/Organization)
3. funder,http://vivoweb.org/ontology/core#informationResourceSupportedBy.http://vivoweb.org/ontology/core#assignedBy
4. funder,suggest some other format

progval:
> However, it is a question whether it is clear enough for people who would like to implement metadata transformation?

Hmm, good point. It doesn't seem to be very readable. (It would also, technically, ambiguous (as `.` is valid in a URI). Fortunately all namespaces involved here end with `#` and not `/` so we can disambiguate, but I'd rather not rely on that)

We could use the `Q{namespace}localpart` syntax from XPath, which would solve the ambiguity.

> It would be better if a crosswalk is expressed in machine-actionable format by using some formal mapping language and tool (for instance [MSCR](https://faircore4eosc.eu/eosc-core-components/metadata-schema-and-crosswalk-registry-mscr), [X3ML toolkit](https://www.ics.forth.gr/x3ml-toolkit), [RML](https://rml.io/specs/rml/), etc.). The CSV approach is not a formal notation for expressing mappings.

I'm not familiar with any of these tools, but I agree on principle. Software Heritage has [a tool](https://gitlab.softwareheritage.org/swh/devel/swh-indexer/-/tree/master/swh/indexer/metadata_dictionary?ref_type=heads) to convert from other formats to Codemeta, but it requires a lot of handholding as it can't use the crosswalks as-is.
But that's probably a subject for a wider discussion (cc @moranegg)

> In this concrete case, the information about funder for a software in VIVO can be found in the following way:
> 
>     * there is an object property http://vivoweb.org/ontology/core#informationResourceSupportedBy linked with a Software instance
> 
>     * range for this object property is a super class of http://vivoweb.org/ontology/core#Grant to ensure you can express that a software might be supported by some community, piece of equipment, grant, etc. Only in the case the object type is http://vivoweb.org/ontology/core#Grant, there might be http://vivoweb.org/ontology/core#assignedBy object property containing information about an Organization Unit representing a Funder.
> 
> 
> Therefore, in this scenario I believe information about range URI might be useful.

I see, these are good points.

> Otherwise, I believe a potential implementer of the crosswalk transformation will have to investigate the VIVO ontology in more details.

This is what happens for other crosswalks in practice, sadly. But it seems to be particularly relevant here.

> However, please select which one of the following four approaches we should follow in our PR, and we will update our PR in accordance with that decision:
> 
>     1. funder,informationResourceSupportedBy.assignedBy (property URI - http://vivoweb.org/ontology/core#informationResourceSupportedBy; http://vivoweb.org/ontology/core#assignedBy; range URI - http://vivoweb.org/ontology/core#Grant; http://xmlns.com/foaf/0.1/Organization)
> 
>     2. funder,http://vivoweb.org/ontology/core#informationResourceSupportedBy.http://vivoweb.org/ontology/core#assignedBy (range URI - http://vivoweb.org/ontology/core#Grant; http://xmlns.com/foaf/0.1/Organization)
> 
>     3. funder,http://vivoweb.org/ontology/core#informationResourceSupportedBy.http://vivoweb.org/ontology/core#assignedBy
> 
>     4. funder,suggest some other format

I'd like to hear other people's preferences/ideas

0hmthu (approved):

dgarijo:
Thanks Moranne,
I will review this PR when time allows (hopefully soon). As for the mapping format, it is true that moving towards sssom seems like the best approach. However, there are cases where we don't have the target namespaces, we have to consider those.
I really like the template Jael created here for FAIR4ML https://docs.google.com/spreadsheets/d/1Z1-GS_omE1mo2-WzWrtENslKd8tUkKqmqP1AsziF4uo/edit?usp=sharing, which is not very machine actionable, but at least contains provenance on the rationale of each decision.

moranegg:
@dgarijo let me know if you have any comments on this PR.

moranegg:
@chenejac can you remove the explanations in ( )
We are trying to improve the mappings with the SSSOM methodology, but in this set, it will be inserted to a very large .csv and all explanations should be stripped from the file to get merged into v3.1.

For more info on mapping evolution:
https://github.com/codemeta/codemeta/blob/sssom-methodology/crosswalks-sssom/README.md

Thank you for the understanding.

chenejac:
> @chenejac can you remove the explanations in ( ) We are trying to improve the mappings with the SSSOM methodology, but in this set, it will be inserted to a very large .csv and all explanations should be stripped from the file to get merged into v3.1.
> 
> For more info on mapping evolution: https://github.com/codemeta/codemeta/blob/sssom-methodology/crosswalks-sssom/README.md
> 
> Thank you for the understanding.

Sure, no problem. It has been done in this [commit](https://github.com/codemeta/codemeta/pull/366/commits/52b28f9b11aa068b1d1636ebe3571c983bfe9b7a)

moranegg (approved):

PR #370, opened 2024-12-20
add Usgsmodelcat mapping

hsu000001:
Initial mapping. Additional USGS Model Catalog data model details are at https://github.com/DOI-USGS/modelcatalog-tam

Unsure of how to map multiple relevant fields to one field (ex: code repository), and how to add conditional statements (ex: author).


hsu000001:
Updated to delimit multiple fields in the mapping to a slash surrounded by spaces

progval:
Do you have a list of definitions for the USGS Model Catalog? I see the schema definition, but it doesn't explain the semantics of terms.

And could you remove the changes to `aggregate.py`?

hsu000001:
The definitions can be seen in the body of here:
https://github.com/DOI-USGS/modelcatalog-tam/blob/main/metadata_schema/BasicProfileSchema.py

There are some other definitions in additional files in that directory, e.g.,
https://github.com/DOI-USGS/modelcatalog-tam/blob/main/metadata_schema/PersonSchema.py

I have attempted to remove the changes from aggregate.py, I think it was only the crosswalk.csv, which I have reverted.

crosswalks/usgs-modelcatalog.csv
```diff
@@ -0,0 +1,74 @@
+Property,usgs-modelcatalog
+codeRepository,software.usgs_software_gitlab_release_urls / software.external_software_repository_urls
```
progval, line 2: `usgs_software_gitlab_release_urls` sounds like it would point to tarballs/zipballs instead of a VCS repository. Isn't `installUrl` a more relevant mapping for this term? Though examples in https://github.com/DOI-USGS/modelcatalog-tam/blob/6582b097d27a1fd461131cf0c74f19937f7abd4a/tests/test_basicmetadata.py look like it's webpages. 

`external_software_repository_urls` also seems to point to webpages in practice, eg. https://github.com/DOI-USGS/modelcatalog-tam/blob/6582b097d27a1fd461131cf0c74f19937f7abd4a/tests/test_basicmetadata.py#L124

crosswalks/usgs-modelcatalog.csv
```diff
@@ -0,0 +1,74 @@
...
+operatingSystem,
+permissions,
+processorRequirements,
+releaseNotes,
+softwareHelp,
+softwareRequirements,
+softwareVersion,version
+storageRequirements,
+supportingData,
+author,person.name when person_type=Author
```
progval, line 22:
```suggestion
author,person when person_type=Author
```

Codemeta's `author`, like usgs-modelcatalog's `person`,  points to an object; and then we map that object's properties individually. `name` is already mapped below.

crosswalks/usgs-modelcatalog.csv
```diff
@@ -0,0 +1,74 @@
...
+keywords,science_keywords / other_keywords
+license,license
+producer,organization / external_organization_name
+provider,
+publisher,
+sponsor,
+version,version
+isAccessibleForFree,
+isPartOf,related_modelcatalog_assets
+hasPart,related_modelcatalog_assets
```
progval, line 43: Hmm... since `related_modelcatalog_assets` doesn't map to either, it might be best to omit it entirely. (And there is also the possibility that a "related asset" is neither a superset not a part of the model being described, right?

crosswalks/usgs-modelcatalog.csv
```diff
@@ -0,0 +1,74 @@
...
+position,
+description,description
+identifier,identifier.name / identifier.id / identifier.url / identifier.type
```
progval, line 46:
```suggestion
identifier,identifier.url
```

In Codemeta/JSON-LD, `identifier` has to be a URI.

crosswalks/usgs-modelcatalog.csv
```diff
@@ -0,0 +1,74 @@
...
+name,name
+sameAs,
+url,other_links.usgs_website / other_links.weblinks
+relatedLink,other_links.weblinks
```
progval, line 50:
```suggestion
url,other_links.usgs_website
relatedLink,other_links.weblinks
```

no need to duplicate. `url` should be the homepage of the project or something similar, so I'm guessing `relatedLink` is a better match for "other links" (especially with the definition saying "Examples include community forums for the model.")

crosswalks/usgs-modelcatalog.csv
```diff
@@ -0,0 +1,74 @@
...
+givenName,
+familyName,
+email,person.email
+affiliation,person.organization
+identifier,person.orcid
+name,person.name
```
progval, line 56:
```suggestion
email,email
affiliation,organization
identifier,orcid
name,name
```

usgs-modelcatalog's `person.email` maps to Codemeta's `author.email` (ditto for the other three), so there is no need to repeat `person.` for each of them.

crosswalks/usgs-modelcatalog.csv
```diff
@@ -0,0 +1,74 @@
...
+maintainer,
+continuousIntegration,
+buildInstructions,
+developmentStatus,
+embargoEndDate,
+funding,
+issueTracker,
+referencePublication,publications.primary_publications / publications.model_citations
+readme,
+hasSourceCode, software.usgs_software_gitlab_release_urls / software.external_software_repository_urls
```
progval, line 73:
```suggestion
hasSourceCode,
```

as in #371, don't map this one as it's used to link SoftwareSourceCode with SoftwareApplication; not to web pages.

crosswalks/usgs-modelcatalog.csv
hsu000001: `usgs_software_gitlab_release_urls` and `external_software_repository_urls` are meant to point to the top page of a public git repository. The confusion may stem from the use of "release" to mean a repository that has gone through proper review and approval processes at the USGS. The published software projects are called "software releases." At the current time, the official release must be on gitlab. However, there are also github repos allowed ("external" because they are not official USGS releases). 
Summary: Those terms both point to webpages/repositories, and not tarballs or zipballs.

crosswalks/usgs-modelcatalog.csv
hsu000001: OK, I wasn't sure how to handle the case that the the thing in the field _might_ be a superset or part of the model, but might not be.  I can omit.

crosswalks/usgs-modelcatalog.csv
hsu000001: I agree mostly, however some complexity is added because some of the models have an official usgs website, while other models do not have an official page on usgs.gov/, so their main page will be in weblinks instead of usgs_website. That should be fine, since all links will be represented, although it bothers me that it is not consistent, due to choices we made for our system.

crosswalks/usgs-modelcatalog.csv
hsu000001: Is there any way to note that person can be of several types and we only want to map these if usgs-modelcatalog's person is of type=author? (as opposed to type=point of contact or type=maintainer)

progval: Not really, but `when person_type=Author` on the `author` line already does it.

hsu000001:
I think that I just have addressed all of the comments or saw an acknowledgment of an explanation. I will await the next step, thanks.

progval (approved):

hsu000001:
Am I able to merge this? Thanks.

PR #371, opened 2024-12-20
add Csdms mapping

hsu000001:
Please see new mapping to csdms.csv

Indicated multiple fields relevant with semicolons.

Did not yet run aggregate.py


progval:
Should this PR replace #370?

Do you have a link to definitions of csdms terms?

> Indicated multiple fields relevant with semicolons.

Could you use a slash surrounded with spaces, instead?

hsu000001:
Hi, #370 and #371 are for two different systems

I will get with csdms folks to find definitions.

I will update both this csdms.csv mapping, and the usgs-modelcat.csv mapping the the  /  for multiple items.

hsu000001:
Now that multiple fields are indicated by " / ", there is one row where the field itself in csdms contains that string, it is on line 54, affiliation,Institute / Organization

where in csdms, the field is "Institute / Organization", but let me check with @kettner

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
+Property,csdms
+codeRepository,Source web address
+programmingLanguage,Programming language / Other Programming language
```
progval, line 3:
```suggestion
programmingLanguage,Programming language / Program language other
```

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
...
+review,
+runtimePlatform,
+targetProduct,
+applicationCategory,module_type
```
progval, line 7: I don't see it on https://csdms.colorado.edu/wiki/Available_model_questionnaire_conditions

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
...
+applicationSubCategory,
+downloadUrl,
+fileSize,
+installUrl,
+memoryRequirements,Memory requirements
+operatingSystem,Supported platforms / Other platform
```
progval, line 13:
```suggestion
operatingSystem,Supported platforms / Supported platforms other
```

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
...
+permissions,
+processorRequirements,Code optimized for (single/multiple processors)
```
progval, line 15: Not a good match, "processorRequirements" is defined as " Processor architecture required to run the application (e.g. IA64). "

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
...
+releaseNotes,
+softwareHelp,
+softwareRequirements,
+softwareVersion,
+storageRequirements,
+supportingData,
+author, First name / Last name
```
progval, line 22: "First name / Last name" are already mapped to givenName/familyName, no need to map them to "author" too.

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
...
+citation,
+contributor,
+copyrightHolder,
+copyrightYear,
+dateCreated,Start year development
+dateModified,End year module development
```
progval, line 28: Shouldn't it be "End year development"? Also, are you sure this is a good match?

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
...
+datePublished,
+editor,
+encoding,
+fileFormat,
+funder,
+keywords,Module domain / model keywords
```
progval, line 34:
```suggestion
keywords,ModelDomain / Model keywords
```

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
...
+license,Program license type
```
progval, line 35:
```suggestion
license,Program license type / Program license type other
```

for consistency

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
...
+producer,
+provider,Contact details modeler
```
progval, line 37: I don't see it on https://csdms.colorado.edu/wiki/Available_model_questionnaire_conditions

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
...
+publisher,
+sponsor,
+version,
+isAccessibleForFree,
+isPartOf,Indicate framework if the model is part of a larger framework
```
progval, line 42: That looks like a definition rather than a term

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
...
+hasPart,
+position,
+description,one-line module description / extended module description
```
progval, line 45:
```suggestion
description,One-line model description / Extended description of the model
```

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
...
+identifier,DOI name model / For model version / Year version submitted / Link to DOI file
```
progval, line 46:
```suggestion
identifier,DOI model / DOI assigned to model version / DOI-filelink
```

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
...
+name,Module name
```
progval, line 47: I don't see it on https://csdms.colorado.edu/wiki/Available_model_questionnaire_conditions

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
...
+sameAs,Also known as
```
progval, line 48: Not a good match. `sameAs` is for alternative URIs, `Also known as` is for alternative names.

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
...
+url,Module website
```
progval, line 49:
```suggestion
url,Model website
```

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
...
+relatedLink,
+givenName,First name
+familyName,Last name
+email,Email address
+affiliation,Institute / Organization
```
progval, line 54: I don't see "Organization" in the list of terms

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
...
+hasPart,
+position,
+description,one-line module description / extended module description
+identifier,DOI name model / For model version / Year version submitted / Link to DOI file
+name,Module name
+sameAs,Also known as
+url,Module website
+relatedLink,
+givenName,First name
+familyName,Last name
```
progval, line 52: Not a very good match: in Eastern cultures, the given name is the last name. But this is a recurring issue we have in mapping with some vocabularies and I don't have a good solution.

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
...
+email,Email address
+affiliation,Institute / Organization
+identifier,
+name,
+address,Postal address 1 / Postal address 2
```
progval, line 57:
```suggestion
address,Postal address1 / Postal address2
```

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
...
+reviewAspect,
+reviewBody,
+endDate,
+roleName,
+startDate,
+softwareSuggestions,
+maintainer,
+continuousIntegration,
+buildInstructions,
+developmentStatus,Status development code / Year when 'status development code' was indicated / Does module development still take place?
```
progval, line 67:
```suggestion
developmentStatus,Development still active
```

because it's the only one that matches well, or

```suggestion
developmentStatus,Development still active / Start year development / End year development 
```

if you think they really should all be there

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
...
+embargoEndDate,
+funding,
+issueTracker,
+referencePublication,
+readme,
+hasSourceCode,Source web address
```
progval, line 73:
```suggestion
hasSourceCode,
```

Don't map this one, it's used to link SoftwareSourceCode with SoftwareApplication; not to web pages.

hsu000001: Acknowledging that this was seen and will discuss with my collaborators, though it may be in the new calendar year.

crosswalks/csdms.csv
```diff
@@ -0,0 +1,74 @@
+Property,csdms
+codeRepository,Source web address
```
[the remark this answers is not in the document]
hsu000001: We would prefer to move in the correct direction and start using Source web address according to its definition, and keep the mapping. But if you prefer that we remove the mapping due to practice, let us know.

crosswalks/csdms.csv
hsu000001: If it is not on the URL you mentioned, it are not available for query. 
However it exists in the schema. See screenshot

Screenshot 2025-01-10 at 10 31 25 AM

crosswalks/csdms.csv
hsu000001: OK, I will remove this mapping.

crosswalks/csdms.csv
hsu000001: Updating to "Institute" only. In the human-readable form, it is listed as "Institute / Organization", but in the URL for the list of terms, it is just "Institute"

crosswalks/csdms.csv
hsu000001: OK. I do think that in CSDMS, even though it says "First Name" and "Last Name", my feeling is that researchers of Eastern cultures would put their family name in the "Last name" spot and their given name in the "first name" spot, even if when saying their name, they would have it reversed. Maybe I am too Western-centric. But as a data manager, I would do that (but maybe I am biased by having Eastern origins). So, since there is no solution so far, can we leave it, since most of the time it maps correctly?

crosswalks/csdms.csv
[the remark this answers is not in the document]
progval: I'm fine with it either way, but you should fix it on your side in order to avoid incorrect translations.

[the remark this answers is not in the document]
hsu000001: OK, we will keep this mapping and work to align.

crosswalks/csdms.csv
hsu000001: We went to check and module_type is on the list, example: https://csdms.colorado.edu/wiki/Property:Model_type , we think this is the correct mapping.

hsu000001:
We resolved or responded to all issues and think this mapping is now complete. Look forward to next step.

crosswalks/csdms.csv
progval: This seems to be how the application is structured, rather than what people use the application for like the examples given on https://schema.org/applicationCategory ("Game, Multimedia"). But I don't know what https://schema.org/applicationCategory is used for, so probably not a big deal

dgarijo: I think application category is more for defining domain-specific terms for findability. It could be similar to what in CSDMS you call "Domain" (Terrestial, Hydrology, etc.)

hsu000001:
Hello, just checking on this PR, thanks.

progval (approved):

progval (changes requested):
There are still several comments that you did not address

crosswalks/csdms.csv
hsu000001: Module name is specified before the terms are filled out, there are workarounds to this, but @kettner do you want to explain further, what should we do here? (we discussed this in a meeting today)

crosswalks/csdms.csv
hsu000001: confirmed

crosswalks/csdms.csv
hsu000001: I think we decided that there would not be a direct map to provider as defined "The service provider, service operator, or service performer; the goods producer. Another party (a seller) may offer those services or goods on behalf of the provider. A provider may also serve as the seller. Supersedes carrier." So we will remove this mapping.

crosswalks/csdms.csv
hsu000001: Agree on both points, so this would not be a good match and the mapping will be removed.

crosswalks/csdms.csv
hsu000001: Okay, we will remove this mapping as the duplicate are not needed.

crosswalks/csdms.csv
hsu000001: We propose that there is a default value here that would apply to all of CSDMS that would be the string "Scientific Model" or "Scientific Tool" - how would I indicate that in the mapping?

crosswalks/csdms.csv
hsu000001: OK, we do not have the URLs for that, so we will remove this mapping.

crosswalks/csdms.csv
hsu000001: The correct term would be ModelFramework (example, Landlab), will update

progval:
> CSDMS has a CodeReviewed field, which is a yes/no, and indicates if the model was reviewed in JOSS. This does not seem to fit in any of the existing fields as specified. How might this be implemented, the URL/DOI of the JOSS paper? But there is not currently a field for that in CSDMS I do not think, so just logging this as an idea. Don't think it should be mapped right now.

Indeed.

> "isAccessibleForFree" should be a default "True" for all entries in CSDMS - not sure how to indicate that in the mapping.

There is no way to express that in crosswalks

progval (approved):

progval:
You can get the error by running `python3 scripts/aggregate.py` on your computer

PR #395, opened 2025-04-10
Initial move of terms into fields

tmorrell:
This PR is for discussion. To do the ssom mapping, we want to have explicit descriptions of sub-properties. This is an initial list based on the fields used in the codemeta generator. We leave out role due to the complexity.


progval:
You need to keep crosswalks in sync with this file

PR #404, opened 2025-04-11
Remove position

tmorrell:
I'm suggesting this change for v4.

This removes the codemeta term position. No one has crosswalked this term, and the use of "list" in authors will ensure that field is ordered without having to include a specific position value. It there are other terms where we want to enforce an order, we should set them as "list"


dgarijo:
+1

mbjones:
+1 to @list from me too

mbjones:
+1 to @list from me too

progval:
+1 from me, but please don't include the unrelated `README: Fix links to v3.0 schema` commit

moranegg:
Added `non-breaking` as this doesn't change the vocabulary.

moranegg:
For v4.0 - breaking change

dgarijo:
The PR for updating the develop branch: https://github.com/codemeta/codemeta/pull/470 (blocking progress here)

PR #410, opened 2025-05-27
Proofread SSSOM Methodology

nichtich:
The documentation and mapping files did not fully conform to SSSOM specification (and still don't do) but this commit removes some errors.

### DIFF
--- a/crosswalks-sssom/README.md
+++ b/crosswalks-sssom/README.md
@@ -1,38 +1,42 @@
 # Mapping-methodology
 
-- **Version**: 0.0.1
+- **Version**: 0.0.2
 - **Authors**: Esteban Gonzalez, Daniel Garijo, Morane Gruenpeter, SciCodes working group
-- **Date**: 01-04-2025
+- **Date**: 2025-05-27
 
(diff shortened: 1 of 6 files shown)


dgarijo:
Thanks for proof reading @nichtich.
Just to clarify, which errors does this remove? I see  `comments` being transformed into `comment`, but nothing else besides reformatting.
Btw, the csv are not meant to be fully compliant with sssom. They are a simplification we did because the participants in the hackathon had some confusion with the names of the columns used in sssom. Therefore the csvs are taken by scripts in the repo and are transformed into the sssom compliant files in each subfolder. See for example: https://github.com/codemeta/codemeta/tree/sssom-methodology/crosswalks-sssom/bibtex/bibtex-sssom-outputs . Those are supposed to be sssom compliant. We also introduce some functions for composite mappings which we did not see supported in sssom but we wanted to keep them for documentation purposes.
@esgg does the `comments` to `comment` break the transformation scripts to sssom?

nichtich:
I changed the definition of `subject_source` and `mapping_provider` in README

PR #414, opened 2025-07-08
Remove non-consequential empty line in codemeta.jsonld

yarikoptic:
it makes reader wonder if there is meaning to it, eg. some groupping of which I think none

It was introduced long ago in 34cb8d12a4b2d454856412c8adc60dd313c87fb0


yarikoptic:
ah, coolio. Some time yaml would prevail over json just to gain ability to comment! ;)

progval:
Unfortunately we're stuck with JSON-LD. I'd happily replace it with RDF/XML or Turtle otherwise.

yarikoptic:
have you looked at/considered https://linkml.io/ ? can produce jsonschema from it etc.

PR #441, opened 2025-10-22
Remove invalid relatedSoftware property

Mazztok45:
This property is not part of the CodeMeta schema and has been removed to maintain validity.


progval (approved):

progval:
Please move the term updates to their own PR. these are a different kind of change

Mazztok45:
> Please move the term updates to their own PR. these are a different kind of change

I removed this commit and will recommit again with another PR, as the crosswalk still needs updates.

PR #303, opened 2023-05-26
Add CFF file automation

tmorrell:
Resolves https://github.com/codemeta/codemeta/issues/297, and would be great to include in the v3 release. This leaves the content of the codemeta file the same; I'll create a separate PR with updates to that.

PR #464, opened 2025-12-10
Add CFF file and automation

tmorrell:
This adds automation to generate a CFF file from the codemeta file in the repo, which partially tackles https://github.com/codemeta/codemeta/discussions/463 and will get a citation box to show up on GitHub. It's an updated version of https://github.com/codemeta/codemeta/pull/303

If the community decides to add a Zenodo DOI for citation purposes, this ensures it will show up in the citation on GitHub.

There one change to the codemeta file that is required for codemeta2cff to work, which is to switch orcid to https. I think we should do this anyway, since it's what's recommended by orcid. But I know there has been lots of discussion of http/https in the past so if we need to make a change in codemeta2cff to accept both that might be possible.

### DIFF
--- a/codemeta.json
+++ b/codemeta.json
@@ -187,7 +187,8 @@
       "@type": "Person",
       "givenName": "Thomas",
       "familyName": "Morrell",
-      "email": "tmorrell@library.caltech.edu"
+      "email": "tmorrell@library.caltech.edu",
+      "@id": "https://orcid.org/0000-0001-9266-5146"
     },
     {
       "@type": "Person",


.github/workflows/codemta2cff.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+      - name: Convert CFF
+        uses: caltechlibrary/codemeta2cff@main
+      - name: Commit CFF
+        uses: EndBug/add-and-commit@v9
+        with:
+          message: 'Add updated CITATION.cff from codemeta.json file'
+          add: 'CITATION.cff'
```
progval:
```suggestion
      - name: Commit CFF
        run: "git commit CITATION.cff -m 'Add updated CITATION.cff from codemeta.json file'"
```

Would this work?

tmorrell: Nope, you get 

```
Author identity unknown

*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"
```

We could add that as a step, but that's replicating a lot of what's in the add-and-commut action. What's the opposition to using an external action?

dgarijo:
Looking at this, it looks like the citation in the README does not correspond with the citation proposed in the PR. We have two options:
- Option 1: Lift contributors to become authors in the `codemeta.json file`.
- Option 2: Adapt the translation in the action so contributors are listed as co-authors in the CFF file.

Consensus from today's meeting is option 1 (@tmorrell,@lindangulopez, @aida, @Mazztok45, @dgarijo were present) .

lindangulopez:
@moranegg can you please weigh in?

moranegg:
@dgarijo a quick comment on authors vs contributors: contributors have contributed mappings, while authors have impacted the vocabulary first design and evolution. 

I will not have capacity to review the PR before our Symposium next week.

PR #471, opened 2026-03-04
Update JSON-LD Context: add alternateName

lindangulopez:
@progval - I will group the rest of the request. This is just to show that I'm working on this issue.

### DIFF
--- a/crosswalks/codemeta-V1.csv
+++ b/crosswalks/codemeta-V1.csv
@@ -1,4 +1,5 @@
 Property,codemeta-V1
+alternateName,alternateName
 codeRepository,codeRepository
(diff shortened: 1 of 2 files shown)


lindangulopez:
Hi @progval , the validate-crosswalks CI is failing I think beacuse I need to change codemeta-V1.csv to add alternativeName. Is that right?

dgarijo:
We should probably first add all the changes we need into the JSON (in one or multiple PRs) and then update the crosswalks.

Otherwise this is going to be unmanageable.

dgarijo:
Please take into account that the duplicate entry for name and identifier may be because one is for author name and another one for the software tool name

Issue #447, opened 2025-11-13
make codemeta more easily citable


alee:
CodeMeta currently asks users to cite codemeta using the w3id but Zotero doesn't recognize it and instead we need to laboriously enter it into Zotero or whatever other reference system we're using AFAICT (also: shouldn't there be a citation.cff in this repo?). The shoemaker has no shoes!

Why not tie into Zenodo and generate a citable DOI instead that goes to a release landing page with the release schema?

Issue #446, opened 2025-11-10
list of each codemeta property with its description


lindangulopez:
Hi, I get to a list of all terms at:   https://codemeta.github.io/terms/ 

For v4, should we consider developing documentation that provides a persistent URL and detailed description for each property to improve discoverability and long-term reference?

Kind regards, 
Linda

rebeccaringuette:
@moranegg potential 4.0 issue

Issue #428, opened 2025-09-24
Add funding information in codemeta.json file


moranegg:

tmorrell:
Included in https://github.com/SciCodes/codemeta/tree/target_codemeta_v4

rebeccaringuette:
Related to [https://github.com/codemeta/codemeta/issues/160](https://github.com/codemeta/codemeta/issues/160)

Issue #427, opened 2025-09-24
Addition of alternateName to CodeMeta v4.0


moranegg:
Following the SSSOM mapping discussion today, it would be useful to add of https://schema.org/alternateName to CodeMeta v4.0

progval:
Why not map them both to `http://schema.org/name`?

dgarijo:
Because the full_name is not really the name of the repo. It's the owner + reponame. This was more or less the agreement of the discussion group, although it's always good to hear other arguments.

If we map both to "name" and we transform the metadata from GitHub API to CodeMeta, it would be unclear which name is the "right" one, unless we add additional code to check for "/".

I think you would enjoy the meetings about mappings. Do you want to join the next one? :)

progval:
> Because the full_name is not really the name of the repo.

Okay, that's a fair point. Thanks for clarifying

tmorrell:
Sounds like a reasonable addition. Needs a PR.

PR #486, opened 2026-07-01
CodeMeta v4 main changes

dgarijo:
This PR will be a bit difficult to review. 
Even though I am doing the PR, it reflects the work done by a community of contributors (see the table in the CONTRIBUTORS file) towards achieving https://github.com/codemeta/codemeta/milestone/11 

Summary of contributions:
- Synced codemeta with schema.org v30 (also using https)
- We created a definition file (codemeta_all.json-ld) following the example of schema.org. That way the `/terms` may redirect to that file, while `/context/` may redirect to the context file.
- Added vocabulary metadata in both files.
- Added missing terms from v4 discussions. All terms reflect the content of the discussions in CodeMeta + consensus from contributors. 
- Added a CFF to ease citation (also updated contributors and metadata files)

Summary of issues tagged as v4 addressed: 

|Issue                                                                                                                   |Issue Number|
|------------------------------------------------------------------------------------------------------------------------|------------|
|Proposal to add property relatedSoftware suggested property                                                             |#290         |
|Should CodeMeta include the concept of "mailing list" / "discussion forum" in next release                              |#166         |
|Linking source code to software applications (entrypoints and service endpoints aka SaaS) with regard for interface type|#271         |
|Should codemeta extend rather than subset schema.org?                                                                   |#161         |
|Remove restriction of 'identifier' to URIs                                                                              |#322         |
|citation property - types should be CreativeWork and Text rather than CreativeWork and URL                              |#268         |
|softwareSuggestions on schema.org                                                                                       |#233         |
|developmentStatus expected types                                                                                        |#339         |
|sorry, revisiting the http:// vs https:// in schema.org once more                                                       |#373         |
|Include codemeta definitions on the context file                                                                        |#323         |
|Range of embargoDate                                                                                                    |#338         |
|Add funding information in codemeta.json file                                                                           |#428         |
|list of each codemeta property with its description                                                                     |#446         |
|Improve the description of buildinstructions                                                                            |#416         |
|Expand the range of buildInstructions                                                                                   |#406         |
|codemeta:maintainer is already in schema.org                                                                            |#405         |
|Addition of alternateName to CodeMeta v4.0                                                                              |#427         |
|Clarify use of property encoding                                                                                        |#418         |
|Adopt schema:creditText                                                                                                 |#407         |
|Update codemeta.json for v4 release                                                                                     |#299         |
|Missing important metadata in context.jsonld                                                                            |#340         |
|make codemeta more easily citable · Issue #447                                                                          |#447         |
|Extend documentation property to include different types of documentation                                               |#255         |
|Remove position                                                                                                         |???         |
|Alignment with v29 of schema.org                                                                                        |???         |
|Citation instructions. A new codemeta property needed?                                                                  |#408         |

The following issues have been postponed, either because they are developer documentation (which is outside the repo) or because they are crosswalk related:

|Issue                                                                                                                   |Issue Number|
-- | --
Describing relationships between research artifacts and code - exploration | #411
Need for controlled vocabularies | #358
Composer crosswalk | #21
Can I add an ror.org id for an organisation? · Issue #308 | #308
"Full example" from user guide is 404 not found · Issue #276 | #276
https://github.com/codemeta/codemeta/issues/288 | #288

Once these changes are accepted, we will move towards v4.1: update other doc, crosswalks, scripts, etc.


dgarijo:
The validation of the checks will likely fail because the changes. We will revisit this in v4.1
I selected @moranegg and @progval per @lindangulopez's request, but please bring in as many people you think should review. Thanks in advance

README.md
moranegg, line 16: Shouldn't we anticipate and add the v4.0 in this README?

moranegg:
Thanks @dgarijo and all!
This looks very nice. I will need time to do the promised high-level review due to the size of this PR.

If anyone from the @codemeta/codemeta-pmc is available to comment, that would be helpful.

meldra:
The problem described in #276 seems to no longer be the case, although the suggested change is not the one that resolved it. Perhaps this issue be closed instead of postponed?

properties_description.csv
```diff
@@ -24,13 +24,15 @@ schema:CreativeWork,citation,CreativeWork or Text,"A citation or reference to an
...
 schema:CreativeWork,datePublished,Date,Date of first broadcast/publication.
 schema:CreativeWork,editor,Person,Specifies the Person who edited the CreativeWork.
 schema:CreativeWork,encoding,MediaObject,A media object that encodes this CreativeWork. This property is a synonym for associatedMedia. Supersedes encodings.
 schema:CreativeWork,fileFormat,Text or URL,"Media type, typically MIME format (see IANA site) of the content e.g. application/zip of a SoftwareApplication binary. In cases where a CreativeWork has several media type representations, 'encoding' can be used to indicate each MediaObject alongside particular fileFormat information. Unregistered or niche file formats can be indicated instead via the most appropriate URL, e.g. defining Web page or a Wikipedia entry."
 schema:CreativeWork,funder,Organization or Person,A person or organization that supports (sponsors) something through some kind of financial contribution.
+schema:CreativeWork,funding,Grant,A Grant that directly or indirectly provide funding or sponsorship for this item
```
meldra: Either "A Grant" should be "Grants", or "provide" should be "provides".

dgarijo: This is taken from https://schema.org/funding There they have it as a link, but I chose not to edit their definition

dgarijo:
> The problem described in #276 seems to no longer be the case, although the suggested change is not the one that resolved it. Perhaps this issue be closed instead of postponed?

By postponed I mean that are not addressed in this PR. But since it's fixed, I went ahead and closed it :)

README.md
dgarijo: Ok, I will add it to the readme

properties_description.csv
dgarijo: I fixed it in any case.

README.md
dgarijo: done

dgarijo:
@moranegg suggested edits are incorporated. Anything else missing?

moranegg:
Thanks for putting this together!

The v4 changes are well organized. Even though this PR is massive and not easy to review. Thank you for your patience and for your group efforts to get this out.

I have doubts about splitting the full term definitions out into `codemeta_all.jsonld` following the schema.org pattern. Since `codemeta_all.jsonld` duplicates the context block, the two need to stay in sync manually.

### Issues with new file codemeta_all.json
What is the advantage of adding the file and having the structure with `@graph` using `rdf:Property`, `rdfs:comment`, and `rdfs:label`?

I'm not against it, but this will require syncing.

In this PR, I've found divergence with the `codeRepository` / `fileFormat`, and undeclared `rdf:` / `rdfs:` prefixes in `codemeta_all.jsonld` which is worth fixing before the merge.

- `codeRepository` has `"@type": "@id"` in `codemeta_all.jsonld` but not in `codemeta.jsonld`. Since its range is URL-only (unlike `identifier`/`url`/`applicationCategory`, which were intentionally de-typed because their ranges include Text), and sibling URL-only terms like `downloadUrl`/`installUrl` kept their typing, I'd restore it in `codemeta.jsonld`.
- `fileFormat` was removed from `codemeta.jsonld` (replaced by `encodingFormat`) but is still present in `codemeta_all.jsonld`.

### Breaking checks

[Check / validate-crosswalks (pull_request)]
- The checks are breaking because the list of properties is now different in v4.0 (as declared in @dgarijo's comment)
Ok to accept the breaking check and merge.

### Smaller fixes

- `CITATION.cff` line 35: `gil@isi.edi` → `gil@isi.edu`
- `CITATION.cff`: the identifier still points to `https://w3id.org/codemeta/3.1` for the 4.0 release, and there's no `version:` field
- `codemeta.json` is internally inconsistent and requires now syncing with the CITATION.cff. 
- `README.md`: "Eseban González-Guardia" → "Esteban"
- `CONTRIBUTORS.MD` (v4 table): `[Mazztok45]` is missing the `@`;  missing space in `Maxence Azzouz\|`; "Fiz Karlsruhe University" should probably be "FIZ Karlsruhe – Leibniz Institute" (it isn't a university)
- `properties_description.csv`: `citeAsID` range is written "String or URL" while the rest of the file uses "Text"

### Questions / non-blocking observations

- The commit message mentions adding `@container` for contributors **and keywords**, but only `contributor` got `"@container": "@list"` — was keywords intentional to leave out?
- Removing `"@type": "@id"` from `identifier`, `url`, `applicationCategory`, etc. changes round-tripping for existing documents (string values that previously expanded to IRIs become literals). That seems to be the point per #322, but it's a breaking change for consumers and probably deserves an explicit callout in the release notes/migration guide.
- `developmentStatus` keeps `"@type": "@id"`, so plain values like `"active"` (as in the repo's own `codemeta.json`) expand to relative IRIs rather than repostatus.org terms. Given #339 expanded its range to include `DefinedTerm`, the example file might want to model that explicitly.
- The deprecation handling via `schema:supersededBy` (`funding`, `maintainer`, `embargoDate`, `contIntegration`) looks good and consistent. 👍

### Funding information (not a blocker)
I would advise adding the OSCARS project, FAIR-IMPACT, FAIRCORE4EOSC, and OSPO-RADAR Sloan funding to the codemeta.json
Similar to what we have on the website with: https://github.com/codemeta/codemeta.github.io/issues/68

If there are no other "hot potatoes" and we find a solution to the syncing situation, we can merge this.
@mbjones @cboettig feel free to add your take.

Long live CodeMeta v4.0!

dgarijo:
@moranegg thanks, a small explanation:
- codemeta_all.jsonld contains the semantics of the vocabualry (domain and range and definitions of each property). The issue is that I copied the context also there. I think it should not be there, I'll fix that. I'll just keep the prefixes and definitions.
- codemeta.jsonld -> just the context. 
If we keep it this way, we don't need to copy paste across files, except keeping in check that if we add a new term we need to add it to the context and the definition file. We had to do this already with the CSV. With this we can also drop the definitions in the csv, which I kept in this PR. codemeta_all should be the source of all definitions.

We could have kept the definitions in the context, but that makes it heavier and people voted not to have it there.

I will need to allocate some time for the other issues.
- Removing @id in identifier, etc. aligns this with schema.org. If we leave it as @id, then string based identifiers would be represented as URIs, which is not what we may want. Interestingly, in `url` they have type as "@id", which would make all urls as URIs, and I find this odd.

mbjones:
Looks like a great direction to me. We've needed the addressable terms list for a long time, so I think this will help a lot.  My only concerns would be around backwards compatibility, such that older codemeta documents in the wild still work, and don't change their meaning. It seems like you've considered that in this refactor. Thanks for the deep work on getting all of those changes done.

moranegg:
@dgarijo I agree that this is a good direction, thank you for pushing it. It is better for readability to keep the two files separate. With that, it will require some syncing at two levels:
(1) between the two jsonld files 
(2) between the jsonld file and the crosswalks

This requires effort and time allocation.

As for the `@id` change, I just wanted to understand. If this aligns with schema.org, I can only support the change.

For the syncing, I suggest keeping this for v4.1, since it will not affect the vocabulary.
But let's not forget it and ship it in 2026.

For the other questions/comments, take your time. I'm away until August 10th and might not be checking very frequently.

dgarijo:
OK, I think I have addressed all the main and minor concerns:
- I removed the context from the definitions file, making it easy to sync in. This fixes the mismatches between the two files
- Added missing rdf namespaces as prefixes.
- Added proposed fixes in CITATION.cff
- Aligned consistency between codemeta.json and CFF.
- Fixed Esteban's name in readme.
- Fixed issues in contributors file
- updated properties_description.csv with the right name

Regarding the questions:
- keywords do not need a container, as the order is not important. 
- Removing the "type" "id" if I remember correctly was already discussed in the corresponding conversations.
- Removed the id from development status to keep it consistent with the definition (it can be text, or id or defined term)
- Added funding info

codemeta.json
```diff
@@ -1,13 +1,13 @@
 {
-  "@context": "https://w3id.org/codemeta/3.0",
+  "@context": "https://w3id.org/codemeta/4",
```
moranegg: Why not 4.0?

codemeta.json
```diff
@@ -1,13 +1,13 @@
...
   "@type": "SoftwareSourceCode",
-  "identifier": "CodeMeta",
+  "identifier": "https://w3id.org/codemeta/4",
   "description": "CodeMeta is a concept vocabulary that can be used to standardize the exchange of software metadata across repositories and organizations.",
   "name": "CodeMeta: Minimal metadata schemas for science software and code, in JSON-LD",
   "codeRepository": "https://github.com/codemeta/codemeta",
   "issueTracker": "https://github.com/codemeta/codemeta/issues",
   "license": "https://spdx.org/licenses/Apache-2.0",
-  "version": "3.1",
+  "version": "4",
```
moranegg: Same question: why not 4.0

codemeta.json
dgarijo: fixed

codemeta.json
dgarijo: fixed

bzg:
I started to split this big PR into small ones.

The first 4 should be straightforward to validate:

- #495 Use https for the schema.org namespace (#373)
- #496 Add aliases for schema.org terms
- #497 Add aliases for `featureList` and `archivedAt`
- #498 Declare `contributor` as an ordered list

Once they are validated, I will move forward with the others. Thanks!

dgarijo:
Once they are validated... by who? Let me know if you need my help

moranegg:
@dgarijo @bzg all, thank you for the support on this.
I'll come back to you on Monday.

moranegg (approved):

moranegg:
@dgarijo I'll be merging this tonight without doing a v4.0 release. Giving the community time before the formal release.

moranegg:
I'm adding my "manual" review here.
[SciCodes-CodeMeta-v4-review.pdf](https://github.com/user-attachments/files/31978045/SciCodes-CodeMeta-v4-review.pdf)

After this merge, we need to take care of the following elements that are hard to review in a big PR:
- position (still appears in description.csv, will be good to check the full coherence with description.csv) 
- ContactPoint and contactPoint

PR #490, opened 2026-07-21
adding `softwareVersion` to the zenodo crosswalk to be converted to `version`

vuillaut:
I'd like to propose adding `softwareVersion` to the Zenodo crosswalk to be converted to `version`.

### DIFF
--- a/crosswalks/Zenodo.csv
+++ b/crosswalks/Zenodo.csv
@@ -16,7 +16,7 @@ processorRequirements,
 releaseNotes,
 softwareHelp,
 softwareRequirements,
-softwareVersion,
+softwareVersion,version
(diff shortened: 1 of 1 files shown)


dgarijo (approved):

dgarijo:
I see it in https://developers.zenodo.org/#representation, although it was not in the previous reference doc we had (https://github.com/zenodo/zenodo/blob/master/zenodo/modules/deposit/jsonschemas/deposits/records/legacyrecord.json). It's a pity that the developer doc is not versioned, as these things change too much at times.
But +1 on the suggestion. Well spotted.

moranegg (approved):

PR #492, opened 2026-07-21
adding crosswalk maintainer -> contributors.ContactPerson for Zenodo

vuillaut:
Hi
I'd like to propose this crosswalk `maintainer` -> `contributors.ContactPerson` for Zenodo

### DIFF
--- a/crosswalks/Zenodo.csv
+++ b/crosswalks/Zenodo.csv
@@ -61,7 +61,7 @@ endDate,
 roleName,
 startDate,
 softwareSuggestions,
-maintainer,
+maintainer,contributors.ContactPerson
 continuousIntegration,
 buildInstructions,
 developmentStatus,


dgarijo:
in v4 we are adding a contactPoint (https://schema.org/contactPoint). I suggest we map it to that. 

I think contactPoint may not be the same as maintainer: the contact point may be something like a mailing list, while the maintainer is a developer in charge of keeping the repo. In many cases it may be the same, but in some others they aren't. 

I suggest we wait on this PR until v4 is out and then make the mapping to contactPoint.

moranegg:
@dgarijo That's a good point.
With that in Zenodo or a Scholarly Repository, the submission contact person is the one maintaining the metadata record and not just a contact email.
I do agree that it is not a 1-to-1 equivalence with the maintainer property.

But there is no `maintainer` property in Zenodo. As much as I could find.

vuillaut:
Hi.
It might not be a 1-to-1 equivalence but imo that's the closest mapping to Zenodo metadata. I let you judge :)

moranegg (approved):

PR #495, opened 2026-09-04
Use https for the schema.org namespace

### LINKED ISSUE #373: sorry, revisiting the http:// vs https:// in schema.org once more
In the context of aligning two national research infrastructures, ODISSEI and CLARIAH, we have hit the following snag. 
One has adopted terms directly from schema.org, using the https:// schema terms.
One has adopted terms directly from codemeta, using the http:// schema terms.

As @proycon highlights various discussions have taken place on this issue: "I remember that this HTTP vs HTTPS issue was an active discussion at CodeMeta. See https://github.com/codemeta/codemeta/pull/248 and https://github.com/codemeta/codemeta/issues/309. The status quo at the time was to stick to HTTP. Even in their latest codemeta 3.0 (which I still have to upgrade to), they're still using HTTP. See https://raw.githubusercontent.com/codemeta/codemeta/3.0/codemeta.jsonld". (see: https://github.com/proycon/codemetapy/issues/52#issuecomment-2633557134).

While I agree with the substantive arguments that moving away from "http://" has no substantial benefit and actually introduces issues, time has moved on and we find that schema.org itself advocates the use of https:// (see: https://github.com/schemaorg/schemaorg/issues/2853 which mentions http:// as deprecated. The schema.org [FAQ](https://schema.org/docs/faq.html#19) is less direct, but clearly states that the future is in the https:// version.

As CODEMETA so wonderfully reuses terms from existing schemas, such as schema.org, I would like to ask for CODEMETA to then also commit to the URI's of said schema and thus move from http:// to https:// uris for the schema.org terms. This would also be congruent with CODEMETA's use of the secure protocol in its own terms/uri's.

Obviously, this issue is not about an 's', nor is it about two dutch infrastructures. However, I feel that more projects may pick up on schema.org first and then find out about CODEMETA later. Or projects may have a need for aligning data first, and then tools later. For CODEMETA to convert to the schema.org recommended version of the URI's, would have great benefits on the global alignment of (research) infrastructures.

bzg:
The `schema` prefix of the CodeMeta context now points to `https://schema.org/` instead of `http://schema.org/`, per the official schema.org recommendation. Two links in the README are also updated.

Closes #373.

Split from #486 (work by @dgarijo and the CodeMeta v4 contributors).

### DIFF
--- a/examples/schema-org-codemeta.json
+++ b/examples/schema-org-codemeta.json
@@ -1,5 +1,15 @@
 {
-    "@context": ["http://schema.org", {"author": {"@id": "schema:author", "@container": "@list"} } ] ,
+    "@context": {
+        "@vocab": "https://schema.org/",
+        "schema": "https://schema.org/",
+        "author": {"@id": "schema:author", "@container": "@list"},
+        "codeRepository": {"@id": "schema:codeRepository", "@type": "@id"},
+        "dateCreated": {"@id": "schema:dateCreated", "@type": "schema:Date"},
+        "dateModified": {"@id": "schema:dateModified", "@type": "schema:Date"},
+        "datePublished": {"@id": "schema:datePublished", "@type": "schema:Date"},
+        "downloadUrl": {"@id": "schema:downloadUrl", "@type": "@id"},
+        "license": {"@id": "schema:license", "@type": "@id"}
+    },
     "name": "Generate CodeMeta Metadata for R Packages",
     "description": "Codemeta defines a 'JSON-LD' format for describing software metadata. This package provides utilities to generate, parse, and modify codemeta.jsonld files automatically for R packages.",
     "identifier": {"@id": "http://dx.doi.org/10.5281/zenodo.XXXX"},
--- a/README.md
+++ b/README.md
@@ -51,5 +51,5 @@ crosswalk table, website, software and other related content.
     - [Code as a research object: updates, prototypes, next steps](http://mozillascience.org/code-as-a-research-object-updates-prototypes-next-steps/)
     - [What else is needed for code reuse](http://mozillascience.org/what-else-is-needed-for-code-reuse/)
     - [JSON-LD for software discovery, reuse and credit](http://www.arfon.org/json-ld-for-software-discovery-reuse-and-credit)
-- [JSON-LD.org](http://json-ld.org/)
-- [Schema.org](http://schema.org/)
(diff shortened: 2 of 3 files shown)


dgarijo:
The change to https is already included in #486 (https://github.com/codemeta/codemeta/pull/486/changes#diff-3fbcc8b672839622712fefa9d80a4bb7d9a1c22b12ebf552f38418e0099ff983)
I suggest not adding it here, as it would probably cause conflicts.

dgarijo:
I suggest considering this PR after #486

examples/schema-org-codemeta.json
```diff
@@ -1,5 +1,15 @@
...
+    "@context": {
+        "@vocab": "https://schema.org/",
+        "schema": "https://schema.org/",
+        "author": {"@id": "schema:author", "@container": "@list"},
+        "codeRepository": {"@id": "schema:codeRepository", "@type": "@id"},
+        "dateCreated": {"@id": "schema:dateCreated", "@type": "schema:Date"},
+        "dateModified": {"@id": "schema:dateModified", "@type": "schema:Date"},
+        "datePublished": {"@id": "schema:datePublished", "@type": "schema:Date"},
+        "downloadUrl": {"@id": "schema:downloadUrl", "@type": "@id"},
+        "license": {"@id": "schema:license", "@type": "@id"}
```
progval, line 11: These changes are not related to the namespace

bzg: What do you suggest?

bzg:
Hi @dgarijo, I've been requested to help reviewing/applying #486 by @moranegg - so I started to split the huge PR into smaller ones, including this one.  Please ping me at bastien.guerry@inria.fr so that we have an opportunity to discuss this, as I don't want to "inflict help".  Thanks!

examples/schema-org-codemeta.json
dgarijo: in theory, just leaving the codemeta namespace should suffice for the context.
This file was not part of the original PR, it's something new

dgarijo:
@bzg thanks for clarifying. I was just confused because initially we had a bunch of PRs in the repo, which we then unified for v4 (many were causing conflicts between them).
As long as I don't need to individually review them again and that the credit for who contributed to the PR is clear, I have no objections if you prefer to split them :)

